PR #647, opened 2026-06-17
feat: add InputSelect composition

isaquebock:
## Summary

Introduce `@aziontech/webkit/inputs/input-select` and its sub-components, composed via a shared injection key:

- `input-select` — root, owns selection state + emits `update:modelValue`.
- `input-select-trigger` — the clickable trigger that opens the listbox.
- `input-select-content` — the floating listbox surface.
- `input-select-group` — optional grouping with header.
- `input-select-option` — individual option, picks selected value via the shared context.

The composition follows the shadcn-vue convention: the consumer controls the structure, the root owns state.

Independent from the other inputs PRs in the series (no shared imports with Label / HelperText / InputText).

## Test plan

- [ ] `pnpm --filter @aziontech/webkit lint` — clean.
- [ ] Storybook → `Webkit/Inputs/InputSelect` → open the trigger, pick an option, value reflects in the controls panel.
- [ ] Keyboard: `Tab` to trigger → `Enter`/`Space` to open → arrows navigate → `Enter` selects → `Esc` closes.

## Known type-check noise

`pnpm --filter @aziontech/webkit type-check` reports the following errors in this PR's scope — they reflect known Vue 3 `Booleanish | undefined` narrowing limitations around `data-*` attrs and an emit overload conflict. Tracking separately; do not block this PR:

- `input-select-content.vue(121,10)` — `Booleanish | undefined` on `data-*` attrs
- `input-select-option.vue(62,6)` — same
- `input-select-trigger.vue(39, 46-48)` — same + `HTMLElement | null` ref typing
- `input-select.vue(85, 106)` — `emit('update:modelValue', value)` overload mismatch


paulo-sobrinho-azion (dismissed):

packages/webkit/package.json
```diff
@@ -104,6 +104,11 @@
     "./textarea": "./src/components/inputs/textarea/textarea.vue",
     "./field-textarea": "./src/components/inputs/field-textarea/field-textarea.vue",
     "./dropdown": "./src/components/inputs/dropdown/dropdown.vue",
+    "./select": "./src/components/inputs/select/index.ts",
```
robsongajunior, line 109: Does not have the selectRoot exported like the another selects.
Any way using select will import all and no make sense use Select with another exported if Select is completed.

robsongajunior: Add a select-root  export.

robsongajunior (changes requested):

robsongajunior (approved):

PR #669, opened 2026-06-24
Fix/checkbox glitch

gabriel-lisboa-azion:
## Overview
Fixes the Webkit checkbox toggle glitch and brings the component, its spec,
and its Storybook story into mutual conformance. Three commits, each scoped
to one concern.

---

### 1. `fix(webkit)` — stop checkbox jumping vertically on toggle
**Context:** the control jumped ~3px vertically every time it was checked/
unchecked.

- **Root cause:** the root `<span>` is `inline-flex` with the default
  `vertical-align: baseline`. While unchecked it has no in-flow children
  (the native input is `absolute`), so its baseline is the box's bottom
  edge; on check, the `v-if` SVG becomes an in-flow flex child and the
  baseline jumps to the icon — shifting the box down on the line.
- **Fix:** `align-middle` on the root makes vertical placement independent
  of content. No effect in flex layouts (e.g. the Disabled row), where
  `vertical-align` is ignored.
- **Verified:** rendered box `top` is identical checked vs unchecked.

_Files:_ `packages/webkit/src/components/inputs/checkbox/checkbox.vue`

---

### 2. `refactor(webkit)` — spec conformance (re-author/approve spec + rewrite component)
**Context:** `.specs/checkbox.md` had drifted from the shipped component and
its checksum no longer validated, so it could not drive `/component-create`.

- Re-authored the spec from the real component (props/events/slots/tokens/
  a11y now match what ships), bumped `spec_version` 2 → 3, and re-approved
  it with a fresh, valid checksum.
- Rewrote the component to the canonical styling pattern the spec's
  Constraints require: all utilities inline on the root, variants via
  `data-*`, dropping the JS class presets (`sharedClasses` / `checkedClasses`
  / `disabledClasses` / `iconClasses` / `rootClasses = computed`) and the
  `interactive-states` preset import.
- Behavior, tokens, and the `align-middle` fix are preserved.
- **Verified:** `validate-spec-compliance` passes (props/events/slots match
  1-to-1); `pnpm --filter webkit lint` clean.

_Files:_ `.specs/checkbox.md`, `packages/webkit/src/components/inputs/checkbox/checkbox.vue`

---

### 3. `docs(storybook)` — rebuild checkbox story per storybook-write spec
**Context:** the story panel and code examples were out of date with the
component.

- `docs.description.component` now injects the spec's Purpose + `## Usage`
  block verbatim, so the Docs code example matches the real API.
- Full `argTypes` for every prop + the `update:modelValue` event, grouped by
  `table.category`.
- Reactive v-model `Template` (local `ref` synced from `args.modelValue`,
  forwards `update:modelValue`) so the panel checkbox actually toggles and
  the Actions panel logs changes.
- Stories limited to the spec's list: Default + a Disabled composite showing
  checked and unchecked.
- No Figma refs, no `argTypesRegex`, no legacy `.args` form.

_Files:_ `apps/storybook/src/stories/webkit/inputs/Checkbox.stories.js`


.specs/checkbox.md
```diff
@@ -2,62 +2,88 @@
...
 
 ## Props
 
 | Prop | Type | Default | Required | JSDoc |
 |---|---|---|---|---|
-| `modelValue` | `string` | `'undefined'` | no | model Value. |
-| `value` | `string` | `'undefined'` | no | value. |
-| `binary` | `boolean` | `false` | no | binary. |
+| `modelValue` | `unknown` | `undefined` | no | Bound value: a boolean in `binary` mode, otherwise the array/value the control reflects. |
+| `value` | `unknown` | `undefined` | no | This checkbox's value, added to or removed from the model array when not `binary`. |
```
robsongajunior, line 41: After update Seems it's updating the type to any and writing 'undefined' inside quotes indicating undefined string.

Type string should default value with '' (empty string).

.specs/checkbox.md
```diff
@@ -2,62 +2,88 @@
...
+| `binary` | `boolean` | `false` | no | Toggles a single boolean instead of collecting `value` into an array. |
 | `disabled` | `boolean` | `false` | no | Disables interaction and applies disabled tokens. |
-| `readonly` | `boolean` | `false` | no | readonly. |
-| `inputId` | `string` | `'undefined'` | no | input Id. |
+| `readonly` | `boolean` | `false` | no | Prevents changes while the control stays focusable. |
+| `inputId` | `string` | `undefined` | no | Forwarded to the native input `id`; pair with an external label's `for`. |
 | `name` | `string` | `undefined` | no | HTML name for form submission. |
-| `tabindex` | `number` | `'undefined'` | no | tabindex. |
```
robsongajunior, line 28: the type if not current default tabindex 0 should be configured to both types `number|undefined` but the case it is writing undefined inside string.

robsongajunior (changes requested):

.specs/checkbox.md
gabriel-lisboa-azion: Fixed in 1b96920: `tabindex` is now typed `number | undefined` with an unquoted `undefined` default. I didn't default it to `0` so the control keeps the browser's native tab ordering unless a consumer opts in.

.specs/checkbox.md
gabriel-lisboa-azion: Addressed in the spec_version 3 rewrite: `value`/`modelValue` are now typed `unknown` (not `any` — our spec Constraints forbid `any`) with an unquoted `undefined` default. I kept `unknown` rather than `string` because the checkbox is polymorphic — the model is a boolean in `binary` mode and an array otherwise — so an empty-string default doesn't apply; `undefined` (no preset value) is correct for a control-only checkbox. The quoted `'undefined'` strings are gone across the table.

robsongajunior (approved):

PR #665, opened 2026-06-24
docs(webkit): codify Show code pattern + flat-import rule

herbert-julio-azion:
## Summary

Documentation/convention PR, rebuilt on top of `dev`. Supersedes #658.

- **storybook-write skill** — documents the copy-paste-ready *Show code* source transform, updated to the current repo format: stories under `apps/storybook/src/stories/components/…`, `title: 'Components/<Category>/<Name>'`, direct `@aziontech/webkit/<name>` imports, and the SFC-detector regex uses `/i` (so generated stories don't trip CodeQL's `js/bad-tag-filter`).
- **New rule `.claude/rules/imports.md`** — codifies the flat public export name: `@aziontech/webkit/<name>`, never a category-prefixed `@aziontech/webkit/<category>/<name>`. The category stays in the source folder and the Storybook title. This is the convention applied across #662/#663/#664 and requested in the #655 review.

## Notes

Replaces #658, branched from `main` (pre-`dev`-reorg).

PR #670, opened 2026-06-24
chore(commands): add /open-pr and /create-branch PR-workflow commands

herbert-julio-azion:
## Summary

Adds two slash commands that encode our PR workflow so anyone gets the same flow:

- **`/create-branch`** — creates a branch off `dev` (never `main`) and **prompts for the related issue** (`ENG-…`), naming the branch `<type>/<ISSUE>-<slug>`.
- **`/open-pr`** — end to end: ensures a `dev`-based branch (runs the create-branch flow if on `dev`/`main`), commits the working changes by context, pushes, and opens the PR against `dev`.

## Conventions baked in

- Base branch is always `dev`; branch from `dev`, not `main`.
- Related issue is required for the branch name.
- Shared docs/rules are flagged to go in a separate PR from code.
- Conventional Commits; no `Co-Authored-By` / attribution footers.

## Notes

`/open-pr` references `.claude/rules/imports.md` (flat exports), which lands in #665.


isaquebock (dismissed):

.claude/commands/open-pr.md
```diff
@@ -0,0 +1,55 @@
...
+- Release rules: `packages/<pkg>/.releaserc` (semantic-release; merges to `dev` cut the version).
+
+## Steps
+
+### 0. Preflight
+- Run `git fetch origin --prune` and `git status`. If there is nothing to commit and nothing unpushed → tell the user there is nothing to open a PR for, and stop.
+
+### 1. Analyze the diff (drives the version bump)
+Run `git diff` (staged + unstaged) and determine, from the change itself:
+- **type** → `feat` (new component / prop / event / slot / public export), `fix` / `hotfix` (bug, a11y, visual regression), `refactor`, `perf`, `docs`, `style`, `chore`, `test`, `ci`, `revert`. This sets the bump: `feat`→minor; `fix`/`hotfix`/`chore`/`docs`/`style`/`refactor`/`perf`→patch; `test`/`ci`/`revert`→no release.
```
robsongajunior, line 22: Should match to release and commit lint.
https://github.com/aziontech/webkit/blob/dev/packages/webkit/.releaserc#L18

If you want add new types, the release file and commit lint should be updated.
The same to @aziontech/theme release config

robsongajunior (changes requested):
Should match to release and commit lint.
https://github.com/aziontech/webkit/blob/dev/packages/webkit/.releaserc#L18

If you want add

.claude/commands/open-pr.md
HerbertJulio: Good catch, addressed in db11d123.

The commit-type set is now **explicit in every `.releaserc` `releaseRules`** (webkit, theme, icons), so it matches `commitlint.config.js` and these commands 1:1:

- added `perf` → `patch`
- added `test` / `ci` / `revert` → `release: false` (no bump)

Before, `perf` only bumped via the conventional-commits preset default and `test`/`ci`/`revert` weren't listed at all — which is exactly the mismatch you flagged.

I also verified in `@semantic-release/commit-analyzer` that a custom rule with `release: false` returns `false` (not `undefined`), so it **suppresses** the default rules. That means `revert` now produces no release, matching what `CONTRIBUTING.md` already documented (previously a well-formed revert would have bumped `patch` via the default `{ revert: true }` rule). The `{ breaking: true }` rule stays last, so breaking commits still resolve to `major`.

Updated `CONTRIBUTING.md` too: dropped the stale "(via preset default)" note on `perf` and clarified that every type is now enumerated in the release config.

robsongajunior (approved):

PR #696, opened 2026-06-30
[NO-ISSUE] feat(icons): add colored brand icons (ai-*-cor) via CSS classes + inline-SVG catalog

gabriel-lisboa-azion:
## Summary

Adds a **colored brand-icon set** (`ai-*-cor`) to `@aziontech/icons`. Multicolor logos (gradients, multiple fills) can't be woff2 font glyphs, so they ship two complementary ways:

- **CSS classes** (`dist/azionicons-color.css`, folded into the `index.css` barrel) — used exactly like a font icon: `<i class="ai ai-google-cor"></i>`. Paints the SVG as a `background-image`, sized to `1em`, and keeps its own palette.
- **Inline-SVG manifest** (`@aziontech/icons/color-catalog`) — raw SVG strings for direct rendering.

17 icons: angular, astro, azuredevops, bitbucket, flutter, gitlab, google, js, kotlin, next, nuxt, react, redwood, solidjs, svelte, ts, vue.

Consumers + a real usage:
- **icons-gallery** and **Storybook `Foundations/Icons`** render the colored set (inline SVG, never recolored) alongside the font icons — searchable and size-adjustable.
- **`sign-up-card`** Google button now shows the colored Google logo (`pi pi-google` → `ai ai-google-cor`).

## Notes

- New package exports: `./azionicons-color` (CSS) and `./color-catalog` (JSON). Colored icons are intentionally excluded from the woff2 font and `catalog.json` (the font-catalog builder is restricted to `ai`/`pi`).
- **No new runtime dependencies. No breaking changes** — additive exports only.
- Includes a small `chore`: migrate `jsconfig.json` (icons + icons-gallery) off TS 6.0-deprecated options (`target=es5`, `baseUrl`, `moduleResolution=node`) → `esnext` / `bundler`.
- `dist/` is gitignored — regenerated by `npm run build` in `packages/icons`.
- The `/include-icons` command documenting this workflow ships in a **separate PR** (shared `.claude/**` tooling).


packages/icons/scripts/build.mjs
```diff
@@ -110,13 +113,15 @@ if (existsSync('./README.md')) {
...
+// ─── Step 7: Generate catalogs (font + colored) ─────────────────────────────
 
-console.log('\n📦 Step 7/7 — Generating dist/catalog.json...\n')
+console.log('\n📦 Step 7/7 — Generating dist/catalog.json + dist/color-catalog.json...\n')
 
 try {
   execSync('node scripts/build-catalog.mjs', { stdio: 'inherit' })
   console.log('  ✔ dist/catalog.json created')
+  execSync('node scripts/build-color-catalog.mjs', { stdio: 'inherit' })
+  execSync('node scripts/build-color-css.mjs', { stdio: 'inherit' })
```
robsongajunior, line 141: Should be before the console.log.
The console.log is the confirmation the end of the process.

robsongajunior (changes requested):

robsongajunior:
The colored SVG payload is duplicated — the steps aren't redundant. The same 17 ai-cor/*.svg files are serialized into two dist artifacts:
- azionicons-color.css (~31.9 KB) — as background-image data-URIs
- color-catalog.json (~32.3 KB) — as inline <svg> strings

Each script (build-color-css.mjs, build-color-catalog.mjs) also re-reads the ai-cor/ directory independently. They filter+sort identically so they can't drift, and each mode is genuinely different (CSS class vs. DOM-inspectable inline SVG) — but a consumer using both (the gallery does) ships the vectors twice. Notably the monochrome catalog.json stores names only (glyph outlines live solely in the woff2), so this overlap is unique to the colored path.

2. How is the SVG used?

Two delivery modes:
- CSS class — .ai-google-cor { background-image: url("data:image/svg+xml,…") }, sized to 1em, used as <i class="ai-google-cor">. The colored class is self-contained (unlike mono, which needs both .ai + .ai-x).
- Inline SVG — color-catalog.json holds raw markup; consumers inject it as real <svg> DOM (e.g. ColoredIcon.vue uses h('span', { innerHTML: svg })), preserving gradients/clipPaths.

3. font-family or base64?

Neither, for colored icons. They are percent-encoded (URL-encoded) UTF-8 SVG data-URIs (data:image/svg+xml,%3Csvg…) — verified base64 appears 0 times in the output. The svgToDataUri() helper uses the mini-svg-data-uri approach (single-quote attrs, escape only % # { } < >), which is smaller than base64 and gzips better. font-family (@font-face woff2 + content:'\fXXX') is the monochrome path only — colored logos can't be single-color font glyphs.

One takeaway worth flagging: if the ~32 KB overlap ever matters, generate azionicons-color.css from color-catalog.json (or serve only the catalog) to give the colored path a single source of truth in dist — right now both are derived independently from source.

robsongajunior (approved):

PR #701, opened 2026-06-30
refactor: rework textarea layout to inline-flex and refresh field-textarea stories

isaquebock:
## Summary

- Refactor `textarea.vue` from absolute-positioned icon slots to an inline-flex row, so left/right icons share spacing tokens with the textarea control and `data-*` attributes on the root wrapper drive disabled/required/invalid styles.
- Filter `class` and `data-testid` out of the textarea passthrough attrs to avoid duplicating them on the inner element.
- Refresh `FieldTextarea` stories: args-driven `Template`, add `Default` and `Disabled` stories, replace the synthetic `Required` story with a submit-driven async demo, and rewrite the docs description block.

## Notes

- No public API change — props, events, and slots are preserved.
- Implied version bump: **patch**.
- Issue: NO-ISSUE.


robsongajunior (changes requested):
- textarea should be rezisable with the 4 default browser params (both, vertical, horizontal, none)
- wrong padding causing visual error

robsongajunior (approved):

PR #705, opened 2026-06-30
[ENG-46310] feat(webkit): add clickable prop and click event to Chip

herbert-julio-azion:
## Summary

Adds an opt-in interactive mode to `Chip` so the chip body can be clicked.

- New prop `clickable` (default `false`). When set, the chip body becomes interactive: `role="button"`, `tabindex="0"`, `cursor-pointer`, a hover/active ghost-layer overlay (`--bg-hover` / `--bg-active`), and a `focus-visible` ring.
- New event `click` (`MouseEvent | KeyboardEvent`). Fires on pointer click, or `Enter` / `Space` while the chip root is focused.
- `clickable` and `removable` compose: the remove button stops propagation, so activating `×` emits `remove` only — never `click`. Keystrokes originating on the remove button do not trigger the chip's activation.
- With `clickable` off, the chip is unchanged (non-interactive).

Spec (`.specs/chip.md`), the Storybook story (new `Clickable` story), and the component are updated 1-to-1; the spec checksum was recomputed.

## Notes

- Version bump: **minor** (`feat`, additive — no breaking change to the existing API).
- `eslint.config.js`: added `KeyboardEvent` to the browser globals — first use in the codebase, required by the new `click` payload type. The repo enumerates DOM globals as components need them (e.g. `MouseEvent` was already listed); without it `no-undef` fails lint.
- No new runtime dependencies; hover/active uses the canonical CSS ghost-layer pattern from DESIGN.md (no animation/positioning lib).
- Relates to ENG-46310.


robsongajunior (changes requested):
Should be added click point and style of chip component.

PR #709, opened 2026-07-01
[NO-ISSUE] feat: rework InputGroup to composition (PrimeVue-style) with <InputGroup.Addon>

isaquebock:
## Summary

Aligns `InputGroup` with the PrimeVue `InputGroup` model: the root only orchestrates borders, `focus-within` ring, and seams; sibling controls (`<Button>`, `<Select>`, `<InputText>`, `<InputGroup.Addon>`) are dropped in as direct children in visual order.

- Spec bumped **v3 → v4** (composition); `<InputGroup.Addon>` documented as the sub-component and compound API.
- Root: flex container with deep selectors that strip descendant `button` / `a` / `[role=combobox]` of their own `border`, `rounded`, and `focus-visible:ring` so the group's border/ring is the single visual authority. Child seams and outer rounded corners applied via Tailwind arbitrary variants.
- **New `<InputGroup.Addon>`** for static-text islands (`https://`, `.com`, `R$`) — `bg-canvas` + `text-muted` + `spacing-md` + `text-label-sm`.
- **Package exports**: `./input-group` → `index.ts` (compound), plus flat `./input-group-root` (tree-shakeable) and `./input-group-addon`.
- `FieldInputGroup` migrated to wrap its own `#left` / `#right` slots in `<InputGroupAddon>` internally — **its public API is unchanged**.
- Stories rewritten: `Default`, `WithAddonLeft/Right`, `BothAddons`, `WithButton`, `WithSelect`, `WithAll`, `Invalid`, `Required`, `Disabled`. Runtime templates use flat sub-component names (`<InputGroupAddon>`, `<SelectTrigger>`, …) to work around Vue's runtime template compiler; `Show code` keeps the compound dot-notation.

## Notes

- Version bump: **minor** (`feat`).
- No new external dependencies.
- The `#left` / `#right` named slots on `InputGroup` v3 are removed. Callers should use `<InputGroup.Addon>` (static text) or drop the interactive component (`<Button>`, `<Select>`) directly as a child. `FieldInputGroup`'s own `#left` / `#right` slots are preserved.

## Test plan

- [ ] `pnpm --filter @aziontech/webkit.dev lint` — passing
- [ ] `pnpm --filter @aziontech/webkit.dev type-check` — passing
- [ ] `pnpm --filter storybook build` — passing
- [ ] Verify in Storybook (dev):
  - [ ] `WithAddonLeft` / `WithAddonRight` / `BothAddons` — addon islands render with `bg-canvas` + muted text
[...]

PR #710, opened 2026-07-01
[NO-ISSUE] refactor: replace Label Required tag with inline * (Required) indicator

isaquebock:
## Summary

Swaps the `Tag` badge used by the `Label` component for the required state (`severity="warning"` `size="small"` `value="Required"`) for a lighter inline text indicator: `<span aria-hidden="true">*</span> (Required)` — the `*` colored `var(--primary)` (the F3652B orange) and the parenthesized `(Required)` in `var(--text-muted)`.

- The asterisk is decorative (`aria-hidden="true"`), so screen readers announce only "(Required)" alongside the label text.
- `Tag` is no longer imported by `label.vue`.
- Spec (`.specs/label.md`) body updated (Purpose, Props JSDoc, Tokens, Accessibility) and checksum refreshed.

## Notes

- Effect propagates automatically to every consumer of `<Label :required>`: `field-text`, `field-password`, `field-checkbox`, `field-radio`, `field-switch`, `field-textarea`, and the upcoming `field-input-group` (opened separately as PR #709).
- No prop API change on `Label` (`required: boolean` behavior is unchanged; only the visual/textual form differs).
- No dependencies added or removed.

## Test plan

- [ ] `pnpm storybook:dev` — open `Components/Inputs/Label` → `Required` story. Verify the `* (Required)` text appears in place of the old orange Tag chip.
- [ ] Same page for every `Field*` component with `required=true` — all inherit the new indicator without further edits.
- [ ] Screen reader smoke test: with a screen reader active, the label reads "<label> (Required)" (no "asterisk" spoken).


herbert-julio-azion (approved):

herbert-julio-azion (changes requested):
Atualize os fields <Label :required>: field-text, field-password, field-checkbox, field-radio, field-switch, field-textarea, para que reflitam o estado atual, já que eles não estão sendo atualizados corretamente.

herbert-julio-azion (approved):

PR #717, opened 2026-07-02
[NO-ISSUE] feat(webkit): adoption toolkit (in-package) + prop-vocabulary guardrails + canonical prop renames

herbert-julio-azion:
## Summary

All inside the existing `@aziontech/webkit` package (**no new packages published**). The branch turns "follow the docs" into "correct by construction, enforced end-to-end":

1. **Adoption toolkit** — any project consumes webkit *correctly by construction*, not just by docs.
2. **Prop-vocabulary guardrails** — one canonical prop/event API across every component, machine-enforced on **create *and* edit**.
3. **Canonical prop renames** (⚠️ **breaking**) — bring existing components onto that vocabulary, and drop redundant events.
4. **Mandatory component tests** — every component ships a `<name>.test.ts` (three-layer architecture from `dev`/#728); existence is now blocking, on save **and** in CI.
5. **`spec-create` split into two modes** — Figma-driven vs. manual (name + category), the manual one deliberately more blocking.

Governance lives in a single registry: **25 standards** (15 `general` that ship to consumers + 10 `webkit`-internal), each paired with a real gate; **11 ESLint rules**; **9 write-time hooks**. An invariant meta-test fails the build if any rule and its gate ever drift apart. Branch is merged up to `dev`.

---

## 1. Adoption toolkit — folded into `@aziontech/webkit` (no new packages)

Everything ships **inside the existing package** as subpaths + bins, over one source of truth:

- **`catalog.json`** — generated from `package.json#exports` + `.specs` (version-locked, regenerated at release). Every tool reads this one file, so nothing drifts from the installed version.
- **FORCE (blocks in editor / pre-commit / CI)** — `@aziontech/webkit/eslint-plugin` (**11 rules**, all `error`: `valid-import-path`, `no-deep-internal-import`, `no-barrel-import`, `no-whole-icon-set-import`, `prefer-tree-shakeable-root`, `no-hardcoded-color`, `no-style-override`, `authoring-standards`, `prefer-define-model`, `no-deprecated-component`, `prefer-webkit-component`) + `@aziontech/webkit/stylelint-config`.
- **GUIDE (assists the AI/dev up front)** — MCP server (`npx -p @aziontech/webkit webkit-mcp`, 10 tools) + a Claude Code bundle, wired by `npx @aziontech/webkit init` / verified by `doctor`. MCP deps are `optionalDependencies`.

## 2. Prop-vocabulary guardrails (`.claude/`)

A single dictionary — `.claude/hooks/_lib/prop-vocabulary.mjs` (documented in `.claude/rules/prop-vocabulary.md`, surfaced to the MCP via `catalog.vocabulary`) — enforced by `validate-spec-compliance` on every `Write`/`Edit`:

- **Prop names:** banned aliases → canonical (`variant`/`appearance`/`intent`/`headerVariant`/`mediaKind`/`cardStyle`→`kind`, `status`→`severity`, `dismissable`/`closeable`→`dismissible`, `heading`/`header`→`title`, `isToggled`→`modelValue`); no `is`/`has` prefix; **positive booleans** (`hidden`→`visible`, …); `size` = `small|medium|large` (order enforced, no `xs/sm/md/lg/xl`).
- **Events:** kebab-case; **no redundant echo** — `<x>-change` is blocked when `update:<x>` exists (use `v-model` / `@update:<x>`; a distinct commit event is a bare `change`).
- **Default drift:** the spec's Default column must match `withDefaults`.
- **`no-style-override`** (consumer eslint): flags `class`/`:class`/`style`/`:style` on a webkit component tag — with an opt-in `style_seam` escape hatch and `options.allow` (slot content / composition is never flagged).
- Parser coverage extended to named `Props` interfaces and `?: () =>` slots.

## 3. Canonical prop renames (⚠️ breaking — major bump)

| Component(s) | Old → New |
|---|---|
| CardPricing | `cardStyle` → `kind` |
| ItemMedia | `mediaKind` → `kind` |
| Table / TableHeader | `headerVariant` → `headerKind` / `variant` → `kind` |
| Switch | `isToggled` → `modelValue`, `type` → `kind`, `isFocused` → `focused` |
| Dialog / Drawer / Popover | `closeable`/`dismissable` → `dismissible` (light-dismiss; `closable` reserved for the X-button — Message/Toast/TabViewItem keep it) |
| OnboardingForm / PickList | `heading`/`header` → `title` |
| Badge / Tag / HelperText / Label | `value` → `label` (slot-fallback text) |
| StatusIndicator | `status` → `severity` (`positive` → `success`) |
| TableSortButton | `hidden` → `visible` (inverted) |

**Redundant echo events removed** (use `v-model` / `@update:*`): `value-change` (CodeBlock/TabView/NavigationMenu), `page-change` (Paginator), `change` (SegmentedButton). Kept: InputNumber `change` (commit-on-blur) and Calendar `month-change` (navigation) — genuinely distinct.

**Deferred (intentional):** ToastItem `type`→`severity` (entangled with the public `toast({type})` store API — one planned break later; canonical set recorded as `neutral|success|info|warning|danger`); BreadcrumbItem keeps `current` (maps to native `aria-current="page"`).

## 4. Mandatory component tests (`testing` standard → `general`)

The three-layer test architecture came in from `dev` (#728): one `<name>.test.ts` per component, **Vitest browser mode** (Playwright Chromium) reusing the Storybook story via `composeStories`, plus `axe-core`. This branch makes it a first-class, **blocking** standard:

- **Existence is mandatory, never grandfathered.** `enforce-test-exists.mjs` (write-time, on `Write` **and** `Edit`/`MultiEdit`) blocks a component `.vue` whose `<name>.test.ts` is missing — creating *or* editing. `check-tests.mjs` (CI, `toolkit` job) fails the PR if any component has no test.
- **Freshness (transition ratchet).** On a PR, a changed component `.vue` must ship its test in the same diff; existing debt is grandfathered in `test-gate-baseline.json` (same philosophy as `check-authoring`), new debt is blocked.
- **`testing` reclassified `webkit` → `general`** (it ships to consumers) and registered in `standards.mjs#enforce[]` (`enforce-test-exists` · `check-tests` · `vitest`); the invariant now knows `check-tests`.
- Added the 5 previously-untested published templates' tests (platform-shell, plan-success, deploy-success, onboarding-form, sign-up-card). The renamed-API fallout from #728's tests was reconciled (17 test files) and 3 real component defects the suite surfaced were fixed.

## 5. `spec-create` split into two modes

`.claude/commands/spec-create.md` + `.claude/skills/spec-create/SKILL.md`:

- **Mode A — Figma-driven** (`--figma <url>`): the frame is the source of truth; extract tokens/regions/states, map to DESIGN.md, ask only where the frame is silent.
- **Mode B — Name + Category** (no `--figma`): no ground truth, so **more blocking** — always asks every section, never infers props/events/tokens, and refuses to write (`BLOCKED:`) rather than leave a mandatory section as `TBD`.

---

## How to test

> Node ≥ 22.

- **Toolkit:** `pnpm --filter @aziontech/webkit test:toolkit` (node:test — eslint rules, MCP, CLI, stylelint).
- **Tests + gates:** `pnpm --filter @aziontech/webkit test` (Vitest browser); `pnpm --filter @aziontech/webkit run test:gate` (existence + freshness); `pnpm --filter @aziontech/webkit run authoring` (construction ratchet); `node --test packages/webkit/test/standards/invariant.test.mjs` (rule⇄gate invariant).
- **Guardrails:** editing a component with a banned prop/alias, a negative boolean, a `<x>-change` echo, default drift, or **without a co-located test** is blocked by the write-time hooks.
- **Consumer:** `no-style-override` flags `class`/`style` on a webkit tag; `npx @aziontech/webkit init` wires a scratch project.

## Follow-ups (not blocking this PR)

- Pre-existing spec↔code animation drift on `dialog` / `table` (from the dev merge) — reconcile separately.
- Pre-existing spec gaps surfaced by the broadened parser (`field-switch` `required`, `menu-item` props).


packages/webkit/scripts/build-catalog.mjs
```diff
@@ -0,0 +1,225 @@
...
+//   - packages/webkit/package.json#exports  (the canonical allowed-import list)
+//   - .specs/*.md                            (per-component API: props/events/slots)
+//
+// Deterministic by design (no timestamps) so a CI drift-check can assert the
+// committed catalog.json matches a fresh build. Re-run at release time via the
+// .releaserc `prepareCmd` so it always tracks the published exports.
+//
+// No external deps — Node built-ins + the repo's own spec parser.
+
+import { readFileSync, writeFileSync, existsSync, readdirSync } from 'node:fs'
```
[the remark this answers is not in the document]
HerbertJulio: Fixed in febc8abb; removed the unused `readdirSync` import from the `node:fs` import list. No behavior change; the catalog regenerates identically (170 exports / 109 components, byte-for-byte same output).

robsongajunior (changes requested):
The file `packages/webkit/test/eslint-plugin/fixtures/catalog.dev.json` should be deleted. The reference `"package": "@aziontech/webkit.dev"` does not exist anymore.

For now It's aligned to only `@aziontech/webkit`.

.claude/agents/_README.md
```diff
@@ -50,14 +50,27 @@ If anything is ambiguous, emit a `BLOCKED: <reason>` line and write no files.
...
+| `emits.md` | Typed `defineEmits`, kebab names, model vs. activation |
+| `event-payloads.md` | Activation events emit `(event, item)` — DOM event first |
+| `slots.md` | Typed `defineSlots`, kebab names, fallback inside the slot |
+| `composables.md` | `readonly` outward, `toValue` args, `onScopeDispose` cleanup; context composable |
+| `root-element.md` | Own the root: polymorphism via `href`, `$attrs` forwarding, minimal `defineExpose` |
+| `component-states.md` | Declared, rendered state surface (loading/empty/error via DS components) |
+| `accessibility.md` | Role, keyboard, focus, `motion-reduce` — owned by the component |
+| `testid.md` | `data-testid` derived `<category>-<name>` / `input-<name>`, overridable |
+| `deprecation.md` | Mark `@deprecated` → one major → remove |
+| `bundle-budget.md` | Per-entry `size-limit`; tree-shaking stays intact |
```
robsongajunior: size-limit has not usage and no cicd configuration,
it is configured only in the `webkit/package.json`

.claude/commands/add-animation.md
```diff
@@ -0,0 +1,22 @@
...
+
+Scaffold a **new** webkit animation in the one correct place — the semantic animation
+catalog — so components never ship a component-local `@keyframes` or an off-catalog
+`animate-[…]`. Triggered by hand, or pointed at by `validate-spec-compliance` when a
+component uses an `animate-*` that is not yet in the catalog.
+
+Run the **add-animation** skill (`.claude/skills/add-animation/SKILL.md`). It:
+
+1. Adds a `.animate-<name>` utility + its `@keyframes` to
+   `packages/theme/src/tokens/semantic/animations.js`, using the `duration` / `curve`
```
robsongajunior, line 11: the `packages/theme/src/tokens/semantic/animations.js`
it 's exported by `@aziontech/theme/animations`

.claude/hooks/validate-tokens.mjs
```diff
@@ -7,87 +7,12 @@
 import { readFileSync } from 'node:fs'
 import { relative, resolve } from 'node:path'
 
+import { TOKEN_CHECKS as VIOLATIONS } from './_lib/token-checks.mjs'
+
 const ROOT = process.cwd()
 const TARGET_PREFIX = 'packages/webkit/src/components/'
 const WIP_PREFIX = 'packages/webkit/src/components/wip/'
```
robsongajunior: we can remove  the `const WIP_PREFIX = 'packages/webkit/src/components/wip/'` and your respective usage, is not being used due moved to the right category.

.specs/dialog.md
```diff
@@ -33,7 +53,7 @@ Layered surface above the page (modal, drawer, menu). Migrated from the existing
 |---|---|---|---|---|
 | `open` | `boolean` | `undefined` | no | Controlled open state. Use with `v-model:open`. |
 | `defaultOpen` | `boolean` | `undefined` | no | Initial open state when uncontrolled. |
-| `closeable` | `boolean` | `undefined` | no | When true, overlay click and Escape close the dialog. |
+| `dismissible` | `boolean` | `undefined` | no | When true, overlay click and Escape close the dialog. |
```
robsongajunior: on this case we are able to set default to false|true,
no reason to `undefined` usage since undefined is not true being like false.

.specs/drawer.md
```diff
@@ -30,7 +50,7 @@ Layered surface above the page (modal, drawer, menu). Migrated from the existing
 |---|---|---|---|---|
 | `open` | `boolean` | `undefined` | no | Controlled open state. Use with `v-model:open`. |
 | `defaultOpen` | `boolean` | `undefined` | no | Initial open state when uncontrolled. |
-| `closeable` | `boolean` | `undefined` | no | When true, overlay click and Escape close the drawer. |
+| `dismissible` | `boolean` | `undefined` | no | When true, overlay click and Escape close the drawer. |
```
robsongajunior: on this case we are able to set default to false|true,
no reason to undefined usage since undefined is not true being like false.

apps/storybook/.storybook/main.js
```diff
@@ -42,8 +42,9 @@ const config = {
       config.plugins.splice(docgenIdx, 0, vue())
     }
 
-    // Dev channel: stories still import from '@aziontech/webkit/*'; redirect to the renamed
-    // workspace package '@aziontech/webkit.dev/*' so no story file needs to change.
+    // The workspace package is @aziontech/webkit again (the .dev dev-channel rename was
```
robsongajunior: remove this comment lines

apps/storybook/src/stories/foundations/Styleguide.mdx
```diff
@@ -0,0 +1,21 @@
+import { Meta, Markdown } from '@storybook/blocks'
+import styleguide from '../../../../../packages/webkit/docs/STYLEGUIDE.md?raw'
```
robsongajunior: Can't cross files from `webkit/theme/icons`.

For an external we need to export in the package.json.
We need maintain control of all external usage files.

Exporting some documentation we will able to use in another projects,
currently work due mono repo usage

packages/webkit/package.json
```diff
@@ -16,6 +16,8 @@
     "clean": "rm -rf dist/",
     "pack:dry": "npm pack --dry-run",
     "pack:check": "! npm pack --dry-run 2>&1 | grep -E '\\.test\\.(ts|js)|src/test/'",
+    "size": "size-limit",
+    "size:why": "size-limit --why",
```
robsongajunior: It is configured only here. Has not cicd usage

packages/webkit/package.json
```diff
@@ -60,25 +92,32 @@
     "@vitest/ui": "^4.1.9",
     "axe-core": "^4.10.3",
     "conventional-changelog-conventionalcommits": "^7.0.0",
+    "eslint": ">=9",
```
robsongajunior: we need keep sync with the root mono repo package.json.

How we was'nt exporting the lints we was using the global configuration.

The eslint, stylelint if contain in the root package.json should be updated both to the same version pattern

packages/webkit/src/eslint-plugin/rules/prefer-webkit-component.js
```diff
@@ -0,0 +1,60 @@
+// Steers consumers OFF foreign UI libraries toward the webkit design system: flags an
+// import from a configured foreign package (default: PrimeVue) and points at webkit.
+// Configurable via options so a team can add its own "don't use X, use webkit" mappings.
+// Catalog-independent (pure import-source matching), so it works even without a catalog.
+
+const DEFAULT_PACKAGES = {
+  // foreign package prefix → human hint
+  primevue:
```
robsongajunior, line 8: what is the relation with Primevue it is complete removed?

robsongajunior (commented):
`packages/webkit/package.json:120` 
the export "./empty-results-block" pointing at src/components/empty-results-block/empty-results-block.vue, a file that does not exist anywhere in the repo, and `catalog.json:1588` ships it as kind: "component". A consumer (or an AI steered by the MCP — getUsageExample('empty-results-block') returns {ok:true} with a "runnable" SFC) imports it, the new valid-import-path rule and validate-references.mjs both accept it because they trust the exports map, and the build fails with module-not-found. The exports key and catalog entry both need removing.

robsongajunior (changes requested):
`packages/webkit/package.json:120` — The PR adds the `export "./empty-results-block"` pointing at `src/components/empty-results-block/empty-results-block.vue`, a file that does not exist anywhere in the repo, and `catalog.json:1588` ships it as kind: "component". A consumer (or an AI steered by the MCP — `getUsageExample('empty-results-block')` returns `{ok:true}` with a "runnable" SFC) imports it, the new `valid-import-path` rule and `validate-references.mjs` both accept it because they trust the exports map, and the build fails with module-not-found. The exports key and catalog entry both need removing.

robsongajunior (changes requested):
## Conflict rule

`packages/webkit/scripts/build-catalog.mjs:121, 267` — Barrel/compound detection tests only `/\/index\.ts$/`, but `tab-view`, `navigation-menu`, `breadcrumb`, and `global-header` export `index.js` barrels. The committed catalog gives them `kind: "other"` and reports the sub-component-retaining barrel itself as the tree-shakeable import. Consequences chain: the MCP's list/search/suggest tools filter `kind === 'component'` so these four components are invisible; `no-style-override` skips their tags (`'other'` is not in `WEBKIT_TAG_KINDS`); and `prefer-tree-shakeable-root` never fires for them.

robsongajunior (changes requested):
`packages/webkit/scripts/check-tests.mjs:57` + .`claude/hooks/enforce-test-exists.mjs:56` — Both test gates only treat `<name>/<name>.vue` as a root, but `spec-compliance-checks.mjs:105` (in this same PR) deliberately also treats `<name>-root.vue` as the root — and `tab-view-root.vue` / `navigation-menu-root.vue` are exactly that shape. Those components are exempt from both the write-time existence block and the CI existence + freshness gates: deleting tab-view.test.ts would pass CI, contradicting testing.md's "existence is mandatory and never grandfathered". The root predicate should live once in `_lib/spec.mjs` and be consumed by all three.

.claude/agents/_README.md
HerbertJulio: You're right; and it was worse than "no CI job": the  binary was never even installed, so the scripts couldn't run anywhere. Fixed for honesty in 5dd5b1c5: removed the dead __TEXT	__DATA	__OBJC	others	dec	hex/ scripts, kept  as the budget of record, updated the  registry entry (now review-gated, with a note saying the automated gate is pending a vue-aware size-limit preset) and this README row now says exactly that. Wiring the real CI job is a follow-up once the tooling supports  entries.

.claude/commands/add-animation.md
HerbertJulio: Good catch; the doc now mentions the public export too: the duration/curve tokens are published as . The internal  path stays in the text because that's the file the skill edits when adding a utility. (5dd5b1c5)

.claude/hooks/validate-tokens.mjs
HerbertJulio: Done;  was dead code (declared, never referenced) since the wip/ zone was removed. Dropped the const and the stale wip/ mention in the header comment. (5dd5b1c5)

.specs/dialog.md
HerbertJulio: Agreed; and it was actually default drift: 's  already sets , , . Spec updated to match the code and the checksum resealed. One deliberate exception:  stays ; it's the controlled half of , where absence means "uncontrolled", not . (5dd5b1c5)

.specs/drawer.md
HerbertJulio: Same fix as dialog: spec now matches 's real defaults; , , , ; checksum resealed.  stays  on purpose (controlled prop, absence = uncontrolled). (5dd5b1c5)

apps/storybook/.storybook/main.js
HerbertJulio: Removed. (5dd5b1c5)

apps/storybook/src/stories/foundations/Styleguide.mdx
HerbertJulio: Fixed properly, as you suggested:  is now a real export of  (exports map entry +  added to the published ), and the MDX imports it via the package specifier instead of crossing packages with a relative path. Also taught  to classify  exports as kind  so the styleguide doesn't pollute the component list. External projects can now consume the doc too. (5dd5b1c5)

packages/webkit/package.json
HerbertJulio: Resolved together with the README thread: the scripts were dead (binary not installed), so they're removed;  stays as the budget of record and the registry now states explicitly that the automated gate is pending; review holds the line meanwhile. (5dd5b1c5)

packages/webkit/package.json
HerbertJulio: Synced; devDependencies now follow the root pattern: ,  (lockfile updated). Note the peerDependencies keep the wide / ranges on purpose: that's the published-package contract for consumers, while the workspace devDeps match the root. (5dd5b1c5)

packages/webkit/src/eslint-plugin/rules/prefer-webkit-component.js
HerbertJulio: PrimeVue is fully removed from webkit itself; no dependency, no styles, no docs references. This rule is about consumer codebases: Console/product repos still import PrimeVue today, and during migration the lint flags each  import and points at the webkit equivalent. PrimeVue is only the default mapping (configurable via options; a team can add or replace packages); it's pure import-source matching, no runtime relation.

robsongajunior (changes requested):
`packages/webkit/src/components/content/accordion/accordion.vue:62-63` — Accordion still emits `value-change` alongside `update:value` (the exact echo pattern this PR removes elsewhere and bans in the vocabulary). It's baselined only in `authoring-baseline.json`, which is read only by the CI ratchet — the write-time `validate-spec-compliance.mjs` hook has no baseline mechanism and `checkEventVocabulary` runs unconditionally, so any future Write/Edit to accordion.vue exits 2. An unrelated one-line fix cannot land without also making the breaking event removal. Meanwhile `catalog.json:2295` still advertises `@value-change` to AI consumers while the same file's `vocabulary` bans it.

robsongajunior:
## About MCP (not required change)

`packages/webkit/scripts/build-catalog.mjs:320` / `packages/webkit/src/mcp/`— `build-catalog.mjs` stamps vocabulary into the catalog "so the MCP surfaces this", and `prop-vocabulary.md` promises it "carries the dictionary to the MCP … so AI-written consumer code uses `kind`, not `variant`" — but zero MCP code reads it. Worse, `mcp/catalog.js#build()` reconstructs the catalog object field-by-field and drops vocabulary at load time, so no tool can even return it accidentally. The promised steering silently never happens.

HerbertJulio:
@robsongajunior; the five review-body findings are addressed (the 10 inline threads were fixed in 5dd5b1c5; these in **73a918bd**):

1. **`fixtures/catalog.dev.json`**; already gone: your `1c9ca35d` (align toolkit to the single package name) removed it; no `@aziontech/webkit.dev` reference remains in the tree.
2. **`./empty-results-block` phantom export**; removed from `exports` and regenerated out of `catalog.json`. You were right about the blast radius: MCP + `valid-import-path` + `validate-references` all trusted the map.
3. **`index.js` barrels**; `build-catalog` now accepts `index.(ts|js)` in all three detection sites; `tab-view`, `navigation-menu`, `breadcrumb` and `global-header` are now real compound components in the catalog (visible to `list/search/suggest`, covered by `no-style-override` and `prefer-tree-shakeable-root`).
4. **Root predicate divergence**; `check-tests.mjs` and `enforce-test-exists.mjs` now match `<name>-root.vue` exactly like `spec-compliance-checks.mjs`; the existence gate went from 75 → **77 roots** (tab-view and navigation-menu are no longer invisible), all with tests, existence + freshness green.
5. **Accordion `value-change` echo**; removed (BREAKING, consistent with the other echo removals in this PR): emit dropped from `accordion.vue`, spec Events row removed + checksum resealed, story argType and test assertions updated, authoring baseline pruned (210 → 209), catalog regenerated.

Local gates on 73a918bd: toolkit 68/68, invariant 9/9, ratchet 0 new, test gates existence+freshness green, accordion/tab-view/navigation-menu suites 56/56.

HerbertJulio:
> About MCP (not required change); build-catalog stamps vocabulary … but zero MCP code reads it; `mcp/catalog.js#build()` drops it at load time.

Confirmed and fixed in **4d51ec94**; you're right that the steering silently never happened: `build()` reconstructed the catalog field-by-field and dropped `vocabulary`, so no tool could return it even by accident. Now `build()` carries it and `list_tokens` (the positive "what to USE" inventory) returns the dictionary in its no-category index; `{ props (canonical + aliases), size, booleanPrefixes, negativeBooleans, events }`; with the tool description updated so the AI actually asks for it. Verified: `listTokens(catalog).vocabulary.props[0].canonical === 'kind'`; toolkit suite 68/68.

robsongajunior (approved):

robsongajunior (changes requested):
# Build Catalog Bug

## 01
**Poblem:**
MCP will render the ESLint plugin as a component. 
The export `./eslint-plugin` classified  Vue component renderable.

What has being commited:
```
"eslint-plugin": {
  "import": "@aziontech/webkit/eslint-plugin",
  "target": "./src/eslint-plugin/index.js",
  "kind": "component",
  "compoundRoot": true,
  "treeShakeableNote": "No tree-shakeable root. Add \"./eslint-plugin-root\" -> the root .vue, ..."
}
```

---

`packages/webkit/scripts/build-catalog.mjs:158-161` the barrel-regex fix now classifies `./eslint-plugin` as a renderable Vue component. `kindFromTarget` has prefix guards for `src/utils|composables|svg|styles|vite` but none for `src/eslint-plugin/`, so its `index.js` hits the broadened `/\/index\.(ts|js)$/` match.

Verified in the committed catalog: `"eslint-plugin": { "kind": "component", "compoundRoot": true, ... }` and live: `MCP list_components` includes it and `getUsageExample('eslint-plugin')` returns `<script setup>import EslintPlugin from '@aziontech/webkit/eslint-plugin'</script><template><EslintPlugin /></template>`.

An AI consumer following the MCP will render the ESLint plugin as a component.

## 02
`packages/webkit/scripts/build-catalog.mjs:276-281`  the `compoundRoots` barrel test runs on all flat subpaths regardless of kind, so `use-controllable`, `use-focus-trap`, `use-placement` (`kind:"composable"`) and `vite `(`kind:"other"`) all carry `compoundRoot: true`, `treeShakeableImport: null`, and a nonsense note telling the maintainer to "Add `./use-focus-trap-root` → the root .vue" for a composable.

Same root cause as #1: the regex was broadened without guarding those two code paths.

robsongajunior (changes requested):
`.claude/hooks/_lib/spec-compliance-checks.mjs:105` — the widened root predicate hard-blocks all future edits to `tab-view-root.vue` and `navigation-menu-root.vue.` Both files now count as roots for the write-time hook, but their pre-existing violations (verified live: `TabViewRoot ≠ TabView`, off-spec props `value/defaultValue`, animation drift; navigation-menu: 7 off-spec props + testid mismatch) are grandfathered only in `authoring-baseline.json`, which the CI ratchet reads and the hook does not — `legacy-components.json` is `[]`. Any Write/Edit to either file exits 2 until every baselined violation is fixed in the same touch. This is the same hook-vs-baseline asymmetry the accordion fix just resolved, recreated on two other components.

robsongajunior (changes requested):
# Components

- `tag.vue:28` — l`abel: undefined` + spec/catalog recording the literal `'undefined'` (violates `props.md` and the PR's own new authoring check).
- `table-sort-button.vue:63` the `:class` ternary remains, and `.specs/table.md:302` still claims `data-visible/data-direction` attributes the component never sets.

robsongajunior (changes requested):
## Test / Lint / MCP

- `packages/webkit/scripts/check-tests.mjs:131` `--update` without `TEST_GATE_BASE` still writes `grandfathered: []`, silently wiping the baseline; no guard was added.
-  `packages/webkit/src/mcp/queries.js:385` `literalDefault` still returns the string `'false'` (truthy), so `getUsageExample('switch')` still emits `<Switch modelValue kind="default" />` — the opposite of the documented default (re-verified live).
-  `packages/webkit/src/eslint-plugin/authoring-checks.js:53` — the `@deprecated` regex still false-flags multi-line JSDoc where the replacement wraps to the next line (re-verified with `node -e`); blocks write-time, CI, and consumer lint.

.claude/hooks/_lib/standards.mjs
```diff
@@ -0,0 +1,296 @@
...
+//               follow; it SHIPS to consuming projects (the consumer guideline + eslint-plugin).
+//   'webkit'  → specific to authoring the webkit design system itself (specs, the exports map,
+//               Storybook, releases, the published package). It does NOT go to consumer projects.
+//
+// enforce[]: { surface, by }
+//   'write-time' → a .claude/hooks/<by>.mjs hook (exit 2)
+//   'lint'       → a packages/webkit/src/eslint-plugin/rules/<by>.js rule (error)
+//   'ci'         → governance.yml job / check-authoring ratchet / commitlint / branch-protection
+//                  / size-limit / vuejs-accessibility
+//   'review'     → mandatory approval — the PR cannot merge without the 2 required approvals.
```
rafaelgarbinatto: Muitos comentarios, faz sentdo ter código em meio a comentários ?

.claude/hooks/_lib/spec.mjs
```diff
@@ -210,16 +249,29 @@ export function constraintsBlockHasCanonicalBullets(body) {
 // ---- Spec lookup from a .vue path ----
 
 /**
- * Given an absolute path to a component .vue file under
- * packages/webkit/src/components/<category>/<name>/<file>.vue,
- * return { category, name, specPath }. Returns null if the path doesn't match.
+ * Given an absolute path to a component .vue file, return { category, name, specPath }.
```
rafaelgarbinatto: Esse tipo de comentário não faz sentido pra mim.

.claude/hooks/_lib/spec-compliance-checks.mjs
```diff
@@ -0,0 +1,245 @@
...
+    const usedAnims = new Set(
+      extractAnimationClasses(vueText)
+        .filter((c) => c.startsWith('animate-') && c !== 'animate-none')
+        .map((c) => c.replace(/^animate-/, ''))
+    )
+    for (const nameOnly of usedAnims) {
+      if (!catalogAnims.has(nameOnly)) {
+        violations.push(
+          `Animation "animate-${nameOnly}" is not in the theme catalog. Run /add-animation ${nameOnly} ` +
+            `to add it to packages/theme/src/tokens/semantic/animations.js + record a Theme gap.`
```
rafaelgarbinatto: Esse tipo de compliance check com regras condicionadas a uma dependência (no casho theme), não me parece fazer sentido, Qual a ideia aqui ?

.claude/hooks/_lib/prop-vocabulary.mjs
```diff
@@ -0,0 +1,199 @@
+// Canonical prop vocabulary — the single source of truth for prop naming across every
```
rafaelgarbinatto: Porque um bloco tão grande de comentário faz sentido?

.claude/docs/DESIGN.md
```diff
@@ -439,12 +439,12 @@ For `animate-popup-scale-in/out`, set `--popup-origin` per instance to match the
 
 The `validate-tokens.mjs` PreToolUse hook enforces these at write time. If a hook blocks: **fix the value, do not work around it.**
 
-- **Colors:** HEX (`#fff`), RGB (`rgb(0,102,255)`), HSL, Tailwind palette names (`bg-blue-500`, `text-gray-700`), PrimeVue color utilities (`text-color`, `surface-50`).
+- **Colors:** HEX (`#fff`), RGB (`rgb(0,102,255)`), HSL, Tailwind palette names (`bg-blue-500`, `text-gray-700`), external/legacy color utilities (`text-color`, `surface-50`).
```
rafaelgarbinatto, line 442: Aqui é faz correto o que mencionei no theme, voce generaliza analisa, e não condiciona mais ao PrimeVue, lá me parece que deveria ser o mesmo caso.

HerbertJulio:
All 7 findings from the four 2026-07-16 reviews are fixed in d44180ed. Point-by-point:

**Build Catalog Bug (review 4709483280)**
- **01: `./eslint-plugin` classified as a renderable component:** `kindFromTarget` now guards `./src/eslint-plugin/`, `./src/mcp/`, and `./src/cli/` first; all classify as `other` before the `index.(ts|js)` branch can match. Regenerated catalog: `"eslint-plugin": { "kind": "other" }`, no `compoundRoot`, and the component count drops 79 → 78 (197 exports unchanged). MCP `list_components` no longer includes it.
- **02: `compoundRoot: true` on composables/vite:** the `compoundRoots` barrel test is now scoped to `^\./src/components/.+/index\.(ts|js)$`, so only real component barrels qualify. Verified in the committed catalog: `use-controllable` / `use-focus-trap` / `use-placement` / `vite` all have `compoundRoot: false`; `tab-view`, `navigation-menu`, `breadcrumb`, `global-header` keep theirs.

**Root predicate hard-blocks `*-root.vue` edits (review 4709637846)**
`validate-spec-compliance.mjs` now reads `packages/webkit/scripts/authoring-baseline.json`; the same ratchet file CI reads, same `<rel>::spec:<violation>` keys; and filters baselined violations before deciding to block. Pre-existing drift on `tab-view-root.vue` / `navigation-menu-root.vue` (already snapshotted in the baseline) no longer blocks unrelated edits; only NEW divergence exits 2. Verified live: piping an Edit event for `mini-button.vue` (a baselined violation) through the hook exits 0 silently.

**Components (review 4709647854)**
- **`tag.vue` `label: undefined`:** default is now `''` in the `.vue`, `.specs/tag.md` (resealed: checksum + `last_updated: 2026-07-16`), the story's argTypes summary, and the regenerated catalog (`"default": "''"`). The icon-only row keeps `undefined` where it is genuinely optional.
- **`table-sort-button.vue` `:class` ternary:** replaced with `:data-visible` / `:data-direction` + `data-[visible=false]:invisible data-[visible=false]:pointer-events-none` Tailwind variants; the component now actually sets the attributes `.specs/table.md:302` documents, and the styling follows `styling.md` (no JS class branching). `table.test.ts` + `tag.test.ts`: 44/44 green.

**Test / Lint / MCP (review 4709694699)**
- **`check-tests --update` baseline wipe:** running `--update` without `TEST_GATE_BASE` now prints `check-tests --update requires TEST_GATE_BASE=<branch>; refusing to wipe the freshness baseline.` and exits 1 (verified live).
- **`literalDefault` truthy `'false'`:** boolean props now resolve `v === true || v === 'true'`; default-on booleans emit the bare attribute, default-off booleans are omitted entirely. Verified live: `getUsageExample('switch')` → `<Switch kind="default" />` (no phantom `modelValue`), `getUsageExample('checkbox')` → `<Checkbox />`.
- **`@deprecated` multi-line false-flag:** the check now consumes JSDoc continuation lines (`* …` until `*/`) as part of the tag's text before testing for emptiness. Verified with `node -e`: `/** @deprecated since 4.2 - use\n * `kind` instead. Removed in 5.0. */` → not flagged; `/** @deprecated */` and `// @deprecated` → still flagged. All three surfaces share the fix (single definition in `authoring-checks.js`).

Full gate run after the fixes: toolkit 68/68, standards invariant green, authoring ratchet 209 known / 0 new, `check-tests` 77/77 components, tag+table browser suites 44/44.

rafaelgarbinatto (changes requested):

.claude/hooks/_lib/spec-compliance-checks.mjs
robsongajunior: @aziontech/webkit/animations

robsongajunior (dismissed):

.claude/hooks/_lib/standards.mjs
HerbertJulio: Enxugado em d02317b3: o header caiu de 20 para 5 linhas e a legenda de `scope`/`enforce[]` agora vive só no `.claude/rules/README.md` (o arquivo aponta para lá).

.claude/hooks/_lib/spec.mjs
HerbertJulio: Comprimido em d02317b3 (8 → 4 linhas). Mantive o essencial porque é contrato de entrada do parser: sem registrar que existem os layouts categorizado e flat (`tag/tag.vue`), o caso flat quebra fácil na próxima manutenção.

.claude/hooks/_lib/spec-compliance-checks.mjs
HerbertJulio: A ideia: a regra tokens.md exige que toda animação venha do catálogo semântico do theme, e o `catalog.json` é a forma legível por máquina disso, commitado e guardado pelo job `catalog-drift` (o PR falha se estiver desatualizado). O ponto válido era o skip silencioso quando o catálogo não lê. Endurecido em d02317b3: componente com `animate-*` agora falha explicitamente pedindo `catalog:build` em vez de pular o check.

.claude/hooks/_lib/prop-vocabulary.mjs
HerbertJulio: Enxugado em d02317b3 (20 → 7 linhas). O racional por nome (`type`/`value` permitidos etc.) ficou só no `.claude/rules/prop-vocabulary.md`; o código só aponta para lá.

.claude/docs/DESIGN.md
HerbertJulio: Feito em d02317b3: `primevue-color` virou `legacy-color-util`, com mensagem e comentários genéricos ("Legacy/external color utility"), o mesmo tratamento do DESIGN.md. O regex segue listando os utilitários concretos que captura (`text-color`, `surface-*`), porque isso o regex precisa mesmo enumerar; só a identidade deixou de ser de marca. Catálogo regenerado junto.

robsongajunior (approved):

rafaelgarbinatto (approved):

PR #719, opened 2026-07-03
[NO-ISSUE] feat(webkit): add Item inline kind, Item.List, and CardBox padded prop

gabriel-lisboa-azion:
## Summary

Extends the **Item** compound and **CardBox** to support the Figma "list of items in a card" and inline-on-page patterns.

- **Item `inline` kind** — new `kind="inline"` value that removes the row's outer padding (transparent surface); gap is preserved.
- **`Item.List`** (new sub-component, `./item-list`) — divided in-card list: full-width bottom dividers between rows (`--border-muted`), no gap. Forces its items to the `default` kind so they never draw a per-item box inside the CardBox.
- **`Item.Group`** — now a gapped (`--spacing-lg`) off-card list that forces its items to `inline`, avoiding a box-in-box effect.
- **CardBox `padded` prop** (optional, default `true`) — pads the content region; set `padded=false` for flush, edge-to-edge content such as an `Item.List`.
- **asChild interactivity** — when `asChild` is set, the merged row (e.g. an anchor) now draws the `focus-visible` ring and hover/active ghost layers, so whole-row links are keyboard- and pointer-legible.

## Notes

- **Non-breaking.** `padded` defaults to `true`, and the three CardBox consumers that pad their own content (`onboarding-form`, `plan-success`, `deploy-success`) were set to `:padded="false"` so they render identically. The CardBox `Default` story drops its inner padding to demonstrate the new default.
- **Behavior change (not API-breaking):** an `Item` placed inside `Item.Group` or `Item.List` has its `kind` coerced (to `inline` / `default` respectively) via a provided container context. Standalone Items are unaffected.
- No new runtime dependencies. Divider/gap/padding all use existing design tokens.
- Storybook: Item stories migrated to the canonical `toSfc` "Show code" source pattern and widened to `w-full`; added `WithList`, `WithListAsChild`, and `Inline` stories. Specs (`item`, `card-box`) updated with refreshed checksums.


.specs/card-box.md
```diff
@@ -22,6 +22,7 @@ Displays content or metadata in the UI. Migrated from the existing implementatio
 | Prop | Type | Default | Required | JSDoc |
 |---|---|---|---|---|
 | `title` | `string` | `undefined` | no | Heading rendered in the header when the `header` slot is empty. |
```
robsongajunior, line 24: "`string` | `undefined`" should "`string` | ''"

robsongajunior (approved):

herbert-julio-azion (approved):

PR #724, opened 2026-07-06
[NO-ISSUE] feat(guardrails): stronger code-standard + animation lint, fix broken gate

herbert-julio-azion:
## Summary

Strengthens the internal guardrails (`.claude/`) so components follow the documented standards automatically, and fixes real bugs found while mapping coverage. Governance-only — separate from the toolkit code PR (#717) per `.claude/rules/git-workflow.md`.

### Fixes (B5)
- **`enforce-component-create.mjs`** called `skillReferencedInTranscript` (undefined; the function is `pipelineReferencedInTranscript`) → it threw and failed open, so the "run the spec pipeline first" gate **never blocked**. Repaired.
- **Dangling `./tokens.md` links** in `no-invention.md` / `migration.md` / `dependencies.md` → repointed to the real catalog doc (`.claude/docs/DESIGN.md § Animations`).

### New enforcement (B1 + inline animations)
- **`validate-spec-compliance`**: blocks forbidden variant-prop names (`variant`/`appearance`/`intent` → `kind`), `is`/`has` boolean prefixes, and camelCase (non-kebab) event names.
- **`validate-tokens`**: blocks JS class presets (`const *Classes = {}/[]/computed`), component-local `<style>` blocks, component-local `@keyframes`, arbitrary `animate-[…]`, and hardcoded `duration`/`ease`/`delay-[…]`.

Both fire only on **new** matches (baseline diff), so existing components are grandfathered. Verified: no false positives on `button` / `input-text` / `badge`; the bad-pattern sample blocks with exit 2; a clean sample passes.

### Not in this PR (follow-ups, noted)
- Animation **catalog cross-check** + **`/add-animation` scaffold** — depend on `tokens.animations` landing in the catalog (in #717), so they follow that merge.
- **Structured usage sections** (When to use / When NOT to use / Related / Best practices) in the spec template + schema + validator, and the **backfill** of existing specs — sequenced together so requiring them can't break existing component authoring.
- `DESIGN.md` animation table intentionally left untouched.


HerbertJulio:
Fechando: consolidando tudo no #717 (sem PR novo). Os commits de governança (gate fix, links, code-standard + inline-animation lint) foram movidos para a branch do #717.

PR #721, opened 2026-07-06
[NO-ISSUE] refactor(storybook): standardize story docs, controls and show-code

herbert-julio-azion:
## Summary

Standardizes **all Storybook stories** against the canonical docs / controls / "Show code" contract (`.claude/rules/storybook-source.md` + the `storybook-write` skill + `Button.stories.js`), so every story reads the same way and its documentation is trustworthy.

- **Show code = runnable SFC.** Every story sets an explicit `parameters.docs.source.code: toSfc(IMPORT, TEMPLATE)`, `docs` stays a plain object literal with a short Purpose lead + `canvas.sourceState`. Copying the "Show code" panel and pasting it renders **exactly** what the canvas shows (correct imports, PascalCase tags, local state embedded when the canvas needs it).
- **Functional, reactive Controls.** One `argTypes` entry per prop/event/slot; the reusable `Template` forwards `args` reactively (never destructured); `v-model` components sync a local `ref` and log their update action in the Actions panel.
- **Composite stories per axis** (`Types` / `Sizes` / …) instead of one-story-per-variant.
- **Tree normalized** to `components/<category>/<name>/` (per `imports.md`); the duplicated Azion logo stories consolidated into a single `AzionLogo` page.

Hardens enforcement so future stories cannot drift:

- `validate-story-source.mjs` now enforces the **full** contract on every write (not only newly introduced violations), adds checks for `source.type: 'dynamic'`, `argTypesRegex`, the legacy CSF2 `Story.args =` form, Figma references, and `args` destructuring; excludes native HTML elements from the lowercase-tag check; and adds an `--all` audit mode.
- Fixes contradictions in `COMPONENT_REQUIREMENTS.md` and the rule doc (drops the "embed the verbatim `## Usage` block" and "destructure handlers off `args`" guidance the hook now blocks).
- **`pnpm storybook:validate-docs`** — a lightweight, opt-in command wrapping the `--all` audit so anyone can check that every component's Storybook docs follow the standard on demand (no CI gate).

## Synced with `dev`

Merged the latest `dev` and resolved the two story conflicts by keeping dev's newer, already-standard-compliant versions:

- **Calendar** — dev's date-range picker rewrite (#707).
- **Item** — dev's `Item.List` + `kind="inline"` content, placed at the PR's normalized `components/content/item/` path with corrected import depth (#719).

[...]

PR #729, opened 2026-07-07
docs(rules): event-payloads standard — activation events emit (event, item)

herbert-julio-azion:
## Summary

Adds `.claude/rules/event-payloads.md`, the shared convention that **activation events emit `(event, item)`** — the DOM `event` is always the first argument, and when the activation targets a specific subject (row, menu action, selected value, tab, chip, crumb) that subject is the second argument. Simple activations with no subject emit `(event)`; model/value events (`update:*`, `value-change`) are unchanged.

This is the **config PR** of a set that aligns the component code to this standard, split by category:
- config (this PR)
- navigation + split-button
- inputs (chip)
- data (table, pick-list)
- templates (deploy-success)

## Notes
- Docs-only; no `packages/*` touched, so semantic-release cuts no version.
- The code PRs implement this convention and reference it.


HerbertJulio:

PR #731, opened 2026-07-07
refactor(webkit)!: chip click/remove emit (event, label)

herbert-julio-azion:
## Summary

Aligns **inputs / chip** to the `(event, item)` event-payloads standard (config PR #729). `click` and `remove` now carry the chip's `label` as the second argument, so a consumer knows which chip was activated.

| Event | Before | After |
|---|---|---|
| `click` | `(event)` | `(event, label)` |
| `remove` | `(event)` | `(event, label)` |

The change is additive (the event was already the first argument), so existing internal consumers that read only the event keep working — this PR is independent of the data PR.

## Verified

## ⚠️ Breaking
Payload gains a second argument; handlers relying on the exact arity should update. Major bump.

### DIFF
--- a/.specs/chip.md
+++ b/.specs/chip.md
@@ -43,8 +43,8 @@ import Chip from '@aziontech/webkit/chip'
 
 | Event | Payload | Notes |
 |---|---|---|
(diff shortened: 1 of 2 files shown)


HerbertJulio:

PR #732, opened 2026-07-07
refactor(webkit)!: data activation events emit (event, item)

herbert-julio-azion:
## Summary

Aligns **data** components to the `(event, item)` event-payloads standard (config PR #729).

| Component | Event | Before | After |
|---|---|---|---|
| table | `row-click` | `(row)` | `(event, row)` |
| table-applied-filters | `remove` | `(filter)` | `(event, filter)` |
| table-applied-filters | `edit` | `(filter)` | `(event, filter)` |
| pick-list | `item-double-click` | `({ list, item, index })` | `(event, { list, item, index })` |

Pick-list's injected `itemDoubleClick` and both list callers updated; `table-applied-filters` captures the event from the composed `Chip` (whose event is already first, so this PR is independent of the chip PR). `edit` accepts `MouseEvent | KeyboardEvent` since the chip is keyboard-activatable.

## Verified

## ⚠️ Breaking
Activation events now emit the DOM `event` first. Major bump.


HerbertJulio:

PR #733, opened 2026-07-07
refactor(webkit)!: deploy-success step-click emits (event, index)

herbert-julio-azion:
## Summary

Aligns **templates / deploy-success** to the `(event, item)` event-payloads standard (config PR #729). `step-click` had the arguments reversed; it now emits the DOM event first.

| Event | Before | After |
|---|---|---|
| `step-click` | `(index, event)` | `(event, index)` |

Self-contained (handler + template binding updated in the same file).

## Verified

## ⚠️ Breaking
Argument order changed; handlers move from `@step-click="(index, event) => …"` to `(event, index) => …`. Major bump.


HerbertJulio:

PR #756, opened 2026-07-17
[NO-ISSUE] test(storybook): cover light theme, tablet and mobile in the visual matrix

robsongajunior:
## Summary

```
16 files modified another are added images
```

The visual regression layer screenshotted every story exactly once — dark theme at 1280×720 — so light mode, tablet, and mobile were never tested, even though light tokens have full parity and the responsive spacing/typography tokens reshape nearly every story below `md`.

Every story is now snapshotted in a **6-mode matrix**: `dark|light × desktop(1280×720)|tablet(768×1024)|mobile(375×667)`, aligned to the theme's token breakpoints. The Storybook toolbar gets the same DS viewports.

- **`.storybook/visual-modes.js`** (new): single source of truth — viewports, theme classes, mode matrix, snapshot naming, `resolveModes` (per-story narrowing + `VISUAL_MODES` env filter). Feeds both the test-runner and `preview.js` (toolbar viewports + theme decorator classes).
- **`test-runner.js`**: per-story mode loop (resize → theme class flip → double-rAF settle → fonts/images settled → screenshot). Failures are collected per mode so one run writes every failing diff. The freeze style now also mirrors `preview.css`'s canvas-transition selectors (they outrank `*` in specificity — a theme flip would otherwise animate 0.4s mid-screenshot). New: explicit image-settle await (load/error + `decode()`) — `networkidle` alone let the remote avatar photo flake both ways.
- **Baseline naming (compat)**: dark-desktop keeps the bare story id, so the 317 committed baselines stay byte-identical; other modes get `--<mode>` suffixes. Verified locally: full matrix = **1,890 snapshots, stable across back-to-back runs (0 diffs, ~2 min)**.
- **Per-story control**: `parameters.visual: false` unchanged; new `visual: { modes: [...] }` narrows the matrix (used only by the 3 foundation token galleries, whose mobile reflow approaches Chromium's 16384px capture ceiling — both themes still covered at desktop).
- **CI**: the governance `visual` job is sharded ×4 (mirrors the Vitest job) via `VISUAL_SHARD`; per-shard diff artifacts. Baseline-regen workflow: timeout 60m + optional `modes` input for partial regeneration. `testTimeout: 30000` in the jest layer (default 5s was sized for one screenshot).
- **Stories**: Breadcrumb's `ResponsiveCollapsed` now uses the DS `mobile` viewport (the old `mobile1` preset no longer exists) — its mobile snapshot finally captures the collapsed ellipsis state.
- **Docs**: `docs/OVERVIEW_TESTS.md` Layer-2 sections updated (matrix, naming, determinism table, sharding, correct workflow filename).

## Notes

- **The `visual` job is expected red on this PR until the Linux baselines land**: next step is dispatching `app-storybook-generate-baseline.yml` on this branch (`update_baselines=true`), then committing the ~1,573 new PNGs here (existing 317 must show zero modifications — that's the compat invariant).
- No release: nothing under `packages/**` changes (`test` type, no bump).
- No new dependencies.
- Root `storybook:test:visual:update:baseline` script now dispatches against the current branch instead of the deleted `test/webkit`.


apps/storybook/.storybook/visual-modes.js
```diff
@@ -0,0 +1,101 @@
...
+        `Unknown visual mode "${mode}". Valid modes: ${Object.keys(MODES).join(', ')}`
+      )
+    }
+  }
+  if (!envFilter) return requested
+  const allow = envFilter
+    .split(',')
+    .map((mode) => mode.trim())
+    .filter(Boolean)
+  return requested.filter((mode) => allow.includes(mode))
```
HerbertJulio, line 88: The `VISUAL_MODES` env filter is never validated against `MODES`, so a typo silently intersects to zero modes: `resolveModes(undefined, 'ligth-desktop')` returns `[]`, `postVisit` hits `modes.length === 0` and the run passes green with no snapshots taken (`resolveModes(undefined, 'dark-desktop,ligth-mobile')` also drops the typo'd mode silently). Since `app-storybook-generate-baseline.yml` feeds its `modes` input straight into this env var, a typo'd dispatch "regenerates" nothing and uploads the unchanged baselines as if it succeeded.

Suggestion: give the env filter the same typo guard the `modes` parameter already has. A valid-but-disjoint filter (e.g. a story narrowed to desktop modes while `VISUAL_MODES=dark-mobile`) still resolves to `[]` and skips, which stays intentional; only unknown names throw.

```suggestion
  const allow = envFilter
    .split(',')
    .map((mode) => mode.trim())
    .filter(Boolean)
  for (const mode of allow) {
    if (!MODES[mode]) {
      throw new Error(
        `Unknown visual mode "${mode}" in VISUAL_MODES. Valid modes: ${Object.keys(MODES).join(', ')}`
      )
    }
  }
  return requested.filter((mode) => allow.includes(mode))
```

apps/storybook/.storybook/preview.js
```diff
@@ -1,46 +1,31 @@
...
 import '@aziontech/webkit/styles/country-flags'
 
-setup((app) => {
-  app.use(PrimeVue, {
-    ripple: false
-  })
+import { withThemeByClassName } from '@storybook/addon-themes'
+import { GLOBALS_UPDATED } from '@storybook/core-events'
+import { addons } from '@storybook/preview-api'
+import { setup } from '@storybook/vue3'
```
HerbertJulio: `setup` became unused when the PrimeVue registration block was removed; it is the only occurrence left in the file. The storybook app sits outside the CI lint scope (the lint job only covers `packages/webkit`), so nothing will flag this dead import.

```suggestion
```

apps/storybook/.storybook/preview.js
```diff
@@ -1,46 +1,31 @@
...
+
 try {
   addons.getChannel().on(GLOBALS_UPDATED, ({ globals }) => {
     if (globals && globals.theme) applyThemeClass(globals.theme)
   })
-} catch {
-  /* headless/test contexts without a channel */
+} catch(error) {
+  console.error(`[ERROR]: ${error}`)
 }
```
HerbertJulio: This catch path is the expected one in headless/test contexts (no addons channel), which is why it was deliberately silent before, with a comment explaining it. Logging it as `[ERROR]` makes every visual test session print a false alarm. Also, `catch(error)` is off the prettier standard (`catch (error)`).

```suggestion
} catch {
  /* headless/test contexts without a channel */
}
```

herbert-julio-azion (changes requested):
Verified two issues on this branch (details inline): the `VISUAL_MODES` env filter silently drops unknown mode names, and `preview.js` keeps a dead `setup` import plus a noisy catch on an expected code path in headless contexts.

apps/storybook/.storybook/visual-modes.js
robsongajunior: done

herbert-julio-azion (approved):

apps/storybook/.storybook/preview.js
```diff
@@ -23,8 +23,8 @@ try {
   addons.getChannel().on(GLOBALS_UPDATED, ({ globals }) => {
     if (globals && globals.theme) applyThemeClass(globals.theme)
   })
-} catch(error) {
-  console.error(`[ERROR]: ${error}`)
+} catch {
+  /* headless/test contexts without a channel */
```
rafaelgarbinatto, line 27: Catch sem saída?

robsongajunior: yes, it is expected to not polute the headless case

rafaelgarbinatto: ok

rafaelgarbinatto (approved):

PR #761, opened 2026-07-19
fix: select/dropdown a11y, z-index and interaction fixes (ENG-46730, ENG-46731, ENG-46739, ENG-46740, ENG-46741)

isaquebock:
## Summary

Fixes five related bugs in the Select/Dropdown surface and introduces
semantic z-index tokens so teleported popovers always sit above modal
and drawer panels — regardless of where the trigger is mounted.

### ENG-46730 — Focus state por teclado não funciona nas opções
**Root cause.** `Dropdown` option classes composed `before:opacity-100`
via array concatenation, which relied on utility ordering after `hover:`
and `focus-visible:` variants — the highlight was inconsistent (and
absent while hovering another sibling).
**Fix.** Highlighted / selected state now flow through `data-highlighted`
and `data-selected` data-attrs on the option button, with Tailwind
`data-[highlighted]:before:opacity-100` variants. Added
`aria-activedescendant` on the combobox trigger and `scrollIntoView` for
the active option so ArrowUp/Down/Home/End nav is visible without
moving DOM focus.

### ENG-46731 — Dropdown renderiza atrás de Drawer/Dialog
**Root cause.** Overlay panels used hardcoded `z-[1000]` (backdrop) and
`z-[1001]` (panel). Popover panels used hardcoded `z-[1100]` — which
technically worked, but was fragile: any consumer setting a higher
z-index on a wrapping container broke the layering, and there was no
single source of truth for what "above overlays" means.
**Fix.** Adds a semantic `z-indices.data.js` token file (built into
`globals.css` alongside spacings / containers) with three canonical
layers:

- `--z-overlay-backdrop: 1000`
- `--z-overlay-panel: 1001`
- `--z-popover: 1100`

Drawer / Dialog overlays + panels and DropdownMenu / Dropdown popovers
were switched to consume these tokens. No `@floating-ui/*` or other
positioning lib was introduced — the fix is CSS variables only (per
`.claude/rules/dependencies.md`).

### ENG-46739 — Clique duplicado ao selecionar opção
**Root cause.** `DropdownMenuItem` emitted BOTH `click` (raw MouseEvent)
and `select` (semantic value) for a single activation. A consumer bound
to `@click` (native + Vue custom) received the handler twice.
**Fix.** Removes the redundant `click` emit; `select` remains the single
semantic activation event. Consumers wanting the raw MouseEvent can
still listen via `@click.native` (Vue 2 pattern) — but in Vue 3 the
correct pattern is `@select`, which now fires exactly once.

### ENG-46740 — DropdownPanel clip content
**Root cause.** Panel had `overflow-hidden rounded-*`, but its inner
scrollable list had no matching `rounded-[inherit]`, so scrollbars and
selected-row surfaces could bleed past the panel's rounded corners on
Safari (which clips absolutely-positioned descendants inconsistently).
**Fix.** Panel is now `flex flex-col isolate` (Safari-safe stacking),
inner list is `min-h-0 flex-1 overflow-y-auto rounded-[inherit]`.
Border radius comes from `var(--shape-button)` — no new tokens
introduced.

### ENG-46741 — Divider deve separar sem padding
**Root cause.** `DropdownMenuSeparator` classes included
`my-[var(--spacing-xxs)]`, which added vertical space around the
divider.
**Fix.** Drops the `my-*` utility; separator now renders as a bare 1px
horizontal rule.

## Files touched

- `packages/theme/src/tokens/semantic/z-indices.data.js` (new)
- `packages/theme/src/scripts/build-tokens.mjs` (wire tokens into build)
- `packages/theme/dist/v3/globals.{css,scss}` + v4 (regenerated)
- `packages/webkit/src/components/inputs/dropdown/dropdown.vue`
- `packages/webkit/src/components/overlay/dialog/dialog-{content,overlay}.vue`
- `packages/webkit/src/components/overlay/drawer/drawer-{content,overlay}.vue`
- `packages/webkit/src/components/overlay/dropdown-menu/dropdown-menu-{content,item}.vue`
- `packages/webkit/src/components/overlay/dropdown-menu/presets/styles.js`

## Test plan

- [ ] Manual — Dropdown story: keyboard ArrowUp/Down/Home/End highlight
      is visible; Enter selects; Escape closes; hover + keyboard don't
      compete.
- [ ] Manual — Drawer/Dialog story with a Dropdown or DropdownMenu
      inside: popover renders ABOVE the panel; z-index tokens visible
      in DevTools.
- [ ] Manual — DropdownMenu: clicking a menu item fires `@select` once,
      and `@click` (if consumer binds it) fires exactly once (native
      only, no duplicate custom emit).
- [ ] Manual — Dropdown panel: scroll the option list; content clipped
      cleanly by the panel's rounded corners on Safari + Chrome.
- [ ] Manual — DropdownMenu: separator renders flush with adjacent
      items, no extra vertical gap.
- [x] `pnpm webkit:lint && pnpm webkit:type-check && pnpm webkit:format:check && pnpm webkit:lint:style && pnpm webkit:type-coverage` — all green locally.


robsongajunior (changes requested):
## 01. The behavior change contradicts both approved specs

Both specs explicitly declare Tab as a close action, and neither is updated in this PR:

- `.specs/select.md:136`: "Tab closes and moves focus to the next form control."
- `.specs/dropdown.md:155`: "Tab closes the panel and lets focus proceed to the next focusable element."

The diff replaces both with a focus trap. Per accessibility.md ("the component implements exactly that map") and no-invention.md, the spec is the contract,  a keyboard-map change must land with the spec change (and re-validation; both specs carry checksums).

On the merits, the specs are also correct per ARIA APG:

- navigation/dropdown is role="menu", the APG menu pattern says Tab closes the menu. The dropdown never had the "unreachable options" bug anyway: it already focuses the first option on open (dropdown.vue:275-282, watch(open) → focusFirstOption()) and restores focus to the trigger on Escape/select. Its Tab change is a pure regression against spec + APG.
- Select is role="combobox"/listbox: APG says Tab closes the popup. A trap is defensible only because this panel can host a search input and footer (dialog-like), but that's a deliberate spec decision, not a patch.


## 02. Rule violations in the trap implementation

- Hand-written trap. accessibility.md is explicit: focus trapping goes through the focus-trap composable (VueUse primitive), "not a hand-written trap." This PR hand-writes two of them.
- Trap without restore. The same rule requires trap and restore-to-trigger. select.vue's setOpen (line 66) never restores focus — so once a user Tabs into the teleported panel, every keyboard exit (Escape, or selecting in single mode) drops focus to <body>. The old Tab-closes path at least left focus somewhere; the trap makes the missing restore much worse. Note navigation/dropdown already has focusTrigger() — the pattern exists in-repo.
- Duplicated logic. The focusable-elements selector + offsetParent filter is copy-pasted into both select-trigger.vue and select-content.vue as two slightly different implementations in the same PR. This belongs in one shared helper (e.g., on the injected context).


## 03. The z-index changes are half-right

- select-content.vue z-50 → z-[1100]: real fix. Dialog/drawer panels sit at z-[1001]/z-[1000]; the teleported select panel at z-50 rendered behind them. 1100 also matches the layer scheme already implied by use-placement.
- dropdown.vue z-[1100] class: a no-op. usePlacement always emits zIndex: '1100' as an inline style (use-placement/index.js:153,166,245), which beats the class. Dead code that plants a second hardcoded copy of the layer constant.
- With this PR there are now three independent hardcoded 1100s (composable default, select class, dropdown no-op class) plus dialog/drawer's 1000/1001. The tokenized --z-* scheme the PR body describes is genuinely the right durable fix — it just isn't in the diff. Ship it or drop it from the description.


## 04. Test coverage: none for the changed behavior

Neither select.test.ts nor dropdown.test.ts mentions Tab at all, and the diff adds no tests. Per testing.md, keyboard behavior changes are exactly what the co-located browser-mode suite exists for, and holding that line is a review responsibility. Needed (whatever the final behavior is): Tab/Shift+Tab from the open trigger, Tab cycling or closing inside the panel, and focus destination after Escape/select.


----

 ## Smaller edge cases

1. All-options-disabled cage: both new handlers do preventDefault() and return when nothing focusable matches — with a panel of only disabled options, Tab is swallowed dead and Escape (which loses focus, see §3) is the only exit. Old behavior closed the panel.
2. Dropdown trap cycles only [role="menuitem"] — a search input in the top slot or footer content is unreachable by the trap, while the select's trap includes inputs/buttons. Two trap semantics in one PR.
3. Implementation details that are fine: ctx.contentId is a plain string ✓; offsetParent filtering works for descendants of a fixed panel ✓; tabindex="-1" options are programmatically focusable and matched by the [role="option"] selector arm ✓; modulo cycling math is correct ✓.

----

**Recommended path**

1. Reconcile body ↔ diff — push the missing commits or rewrite the description to cover only ENG-46730/46731 for select + navigation/dropdown.
2. Fix ENG-46730 the spec-compliant way: focus the first option when the select opens (mirror dropdown.vue's watch(open) → focusFirstOption()), and on Tab inside the panel, close + synchronously refocus the trigger without preventDefault — the browser's default Tab then proceeds from the trigger, satisfying both specs verbatim ("closes and moves focus to the next form control"). Add focus-restore-to-trigger on all close paths.
3. Revert the dropdown Tab change (menu ⇒ Tab closes per APG and its own spec) and the no-op z-[1100] class.
4. If the team deliberately wants a trapping, dialog-like select panel: update .specs/select.md's keyboard map in the same PR, use the VueUse focus-trap composable per accessibility.md, and keep restore-on-close.
5. Keep the select-content z-index fix; ideally land the --z-* tokens the body promises so 1000/1001/1100 have one source of truth.
6. Add Tab + focus-restoration tests to both suites.

PR #772, opened 2026-07-22
[NO-ISSUE] feat: webkit init hardening, toast service auto-mount, and motion token adoption

herbert-julio-azion:
## Summary

Closes the three gaps found while exercising `webkit init` against a fresh consumer project (unstyled app after init, toasts silently not rendering, and a sidebar nobody knew how to animate) — by making init complete the install, moving feature-scoped setup to just-in-time, and making the animation tokens discoverable **and** enforced.

### `webkit init` — completes the install
- Interactive Y/n prompts on a TTY (icons, entry wiring); `--yes`, `--no-icons`, `--no-entry` for scripted runs. Piped/CI runs never prompt.
- **Entry wiring is now a patch, not advice**: `import './webkit.css'` + `import '@aziontech/icons'` are prepended once (idempotent, quote-style aware) to `src/main.*` — the "installed but unstyled" failure can no longer happen.
- **`@aziontech/webkit/styles`** (new export): the Tailwind `@source` now ships inside the package and the generated `webkit.css` imports it by package name — no more `@source '../node_modules/...'` hardcoded in the consumer's CSS (immune to hoisting/workspace layouts).
- `doctor` gains checks: styles/icons imports at the entry, webkit registered with Tailwind, and **toast setup** (warns with the exact fix when `@aziontech/webkit/toast` is imported but no region is wired).

### Toast — standard Vue service, wired just-in-time
- `ToastPlugin.install()` now **mounts the toast region automatically** (`createVNode` + the app's context, torn down on `app.unmount()`, SSR-safe): `app.use(ToastPlugin)` in `main.ts` is the whole setup. Options ride the plugin: `app.use(ToastPlugin, { position: 'top-right' })`. A manually mounted `<Toaster />` stays harmless (the store activates one region).
- Init deliberately does **not** wire it: the component declares its one-time wiring in a new catalog `setup` field (from spec frontmatter), surfaced by the MCP's `get_component`/`get_best_practices`, documented in the shipped bundle, and backstopped by the `doctor` check above.

### Motion — consumers can find and must use the animation tokens
- **Theme**: new lateral-panel animations `animate-slide-in-left/out-left` and `slide-in-right/out-right` (timings from the duration/curve tokens), plus a `useWhen` map documenting every `animate-*` utility next to its definition.
- **Catalog/MCP**: `tokens.animationGuide` (class + timing + use-when per animation) and a new `list_tokens` category `animations`.
- **Guidance**: the panel recipe (Vue `<Transition>` + the catalogued pair + `motion-reduce:` escape) added to the shipped `webkit-motion` rule and `webkit-motion-polish` skill.
- **Storybook**: new **Foundations/Motion** page — every animation with a replayable preview, the duration/curve token tables, and the panel recipe.
- **Enforcement**: new `webkit/no-hardcoded-motion` ESLint rule (arbitrary `duration-[…]`/`ease-[…]`/`animate-[…]`, literal ms / `cubic-bezier()` / `transition: all` in style strings, `animate-*` names not in the catalog, motion classes without a `motion-reduce:` escape) + the stylelint mirror for CSS declarations. Registered in the standards registry (no orphan gate).

## Notes

- Validated end-to-end against a real consumer app via a local registry: init wires the entry by itself, the build emits the component classes and icon fonts, the sidebar animates with `animate-slide-in-left/out-left` (verified in Chromium, including `prefers-reduced-motion` → `animation: none`), and the new lint flags all three violation classes in consumer code.
- Toolkit suite 96/96 (CLI, MCP, eslint-plugin, stylelint, standards invariant), toast browser suite 21/21, Storybook build green, doc-standards ratchet clean.
- `.specs/toast.md` is updated with the auto-mount Service contract (checksum recomputed); `@keyframes` for the new animations are transform-only.
- No new dependencies. No breaking changes: `ToastPlugin.install()` gains behavior (auto-mount); a manual `<Toaster />` still renders no duplicate region — but see the upgrade note below for who owns the region's props.

## Upgrade note — toast auto-mount

If you already install the plugin **and** mount a `<Toaster position="…" />` manually, the plugin's auto-mounted region now registers first and becomes the active one — the props on your manual `<Toaster>` silently stop applying. Move them to the plugin: `app.use(ToastPlugin, { position: '…' })` (and drop the manual `<Toaster>`; keeping it is harmless but inert). Consumers using only one of the two patterns are unaffected.

### DIFF
--- a/packages/theme/src/tokens/primitives/animations/animate.js
+++ b/packages/theme/src/tokens/primitives/animations/animate.js
@@ -29,8 +29,6 @@ export const animate = {
   'highlight-fade': 'highlight ease-in forwards',
   'popup-scale-in': `popupScaleIn ${duration['moderate-01']} ${curve['productive-entrance']}`,
   'popup-scale-out': `popupScaleOut ${duration['fast-02']} ${curve['productive-exit']}`,
-  // Lateral panels (sidebar / drawer): the panel slides from its anchored edge.
-  // Entrance is a touch longer than popups (larger surface); exit is quicker.
   'slide-in-left': `slideInLeft ${duration['moderate-02']} ${curve['productive-entrance']}`,
   'slide-out-left': `slideOutLeft ${duration['moderate-01']} ${curve['productive-exit']}`,
   'slide-in-right': `slideInRight ${duration['moderate-02']} ${curve['productive-entrance']}`,
@@ -39,11 +37,6 @@ export const animate = {
   'progress-indeterminate-short': `progressIndeterminateShort ${duration['slow-04']} ${curve['expressive-entrance']} ${duration['slow-03']} infinite`
 }
 
-/**
- * When to reach for each `animate-*` utility — the guidance half of the catalog,
- * surfaced to consumers by the webkit MCP (`list_tokens`) and the Storybook
- * Foundations/Motion page. Every `animate` key has an entry; keep them in sync.
- */
 export const useWhen = {
   spin: 'Indeterminate circular spinners (loading icons).',
   ping: 'One-off attention ring radiating from a small element (notification dot).',
--- a/packages/theme/src/tokens/primitives/animations/animate.js
+++ b/packages/theme/src/tokens/primitives/animations/animate.js
@@ -29,8 +29,43 @@ export const animate = {
   'highlight-fade': 'highlight ease-in forwards',
   'popup-scale-in': `popupScaleIn ${duration['moderate-01']} ${curve['productive-entrance']}`,
   'popup-scale-out': `popupScaleOut ${duration['fast-02']} ${curve['productive-exit']}`,
+  // Lateral panels (sidebar / drawer): the panel slides from its anchored edge.
+  // Entrance is a touch longer than popups (larger surface); exit is quicker.
+  'slide-in-left': `slideInLeft ${duration['moderate-02']} ${curve['productive-entrance']}`,
+  'slide-out-left': `slideOutLeft ${duration['moderate-01']} ${curve['productive-exit']}`,
+  'slide-in-right': `slideInRight ${duration['moderate-02']} ${curve['productive-entrance']}`,
+  'slide-out-right': `slideOutRight ${duration['moderate-01']} ${curve['productive-exit']}`,
   'progress-indeterminate': `progressIndeterminate ${duration['slow-04']} ${curve['productive-entrance']} infinite`,
   'progress-indeterminate-short': `progressIndeterminateShort ${duration['slow-04']} ${curve['expressive-entrance']} ${duration['slow-03']} infinite`
 }
 
-export default { animate, curve, duration }
+/**
+ * When to reach for each `animate-*` utility — the guidance half of the catalog,
+ * surfaced to consumers by the webkit MCP (`list_tokens`) and the Storybook
+ * Foundations/Motion page. Every `animate` key has an entry; keep them in sync.
+ */
+export const useWhen = {
+  spin: 'Indeterminate circular spinners (loading icons).',
+  ping: 'One-off attention ring radiating from a small element (notification dot).',
+  pulse: 'Skeleton/placeholder opacity pulse while content loads.',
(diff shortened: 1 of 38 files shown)


packages/theme/src/tokens/primitives/animations/animate.js
```diff
@@ -29,8 +29,43 @@ export const animate = {
   'highlight-fade': 'highlight ease-in forwards',
   'popup-scale-in': `popupScaleIn ${duration['moderate-01']} ${curve['productive-entrance']}`,
   'popup-scale-out': `popupScaleOut ${duration['fast-02']} ${curve['productive-exit']}`,
+  // Lateral panels (sidebar / drawer): the panel slides from its anchored edge.
```
isaquebock: remover comentários

packages/theme/src/tokens/primitives/animations/animate.js
```diff
@@ -29,8 +29,43 @@ export const animate = {
...
+  'slide-out-left': `slideOutLeft ${duration['moderate-01']} ${curve['productive-exit']}`,
+  'slide-in-right': `slideInRight ${duration['moderate-02']} ${curve['productive-entrance']}`,
+  'slide-out-right': `slideOutRight ${duration['moderate-01']} ${curve['productive-exit']}`,
   'progress-indeterminate': `progressIndeterminate ${duration['slow-04']} ${curve['productive-entrance']} infinite`,
   'progress-indeterminate-short': `progressIndeterminateShort ${duration['slow-04']} ${curve['expressive-entrance']} ${duration['slow-03']} infinite`
 }
 
-export default { animate, curve, duration }
+/**
+ * When to reach for each `animate-*` utility — the guidance half of the catalog,
```
isaquebock: remover comentários

packages/theme/src/tokens/primitives/animations/animate.js
HerbertJulio: Removido. A orientação já vive no mapa useWhen logo abaixo (é o que alimenta o MCP e a página Foundations/Motion), e as demais entradas do mapa não levam comentário.

packages/theme/src/tokens/primitives/animations/animate.js
HerbertJulio: Removido. A única informação que o bloco carregava de fato era o invariante (toda chave de animate precisa de entrada no useWhen); isso virou teste em packages/webkit/test/mcp/queries.test.mjs, que agora falha se uma animação nova entrar sem guidance.

isaquebock (dismissed):

robsongajunior (changes requested):
1. `.specs/_schema.json` was not updated for the new `setup`: frontmatter key. The schema declares `additionalProperties: false` and its property list (`name`, `category`, …, `style_seam`, `created`, `last_updated`) does not include `setup` — yet `.specs/toast.md` now carries it and `build-catalog.mjs` reads it. Nothing fails today because the enforced hook path (spec.mjs#validateFrontmatter, used by enforce-spec-exists) is a permissive hand-rolled mirror — but that mirror's comment says it "mirrors _schema.json", and the spec-validate flow validates against the schema, so the next re-validation of toast (or any spec adopting `setup`) will reject it. When `style_seam` was introduced, the schema was updated in the same change; setup needs the same one-line addition (plus a description) in this PR.

2. Auto-mount changes behavior for consumers who already do `app.use(ToastPlugin)` and mount `<Toaster position="…"/>` manually. The plugin's region now mounts synchronously at `use()` — before the app (and the manual Toaster) mounts — and the store's registry activates the first region. So after upgrade, the plugin's region (plugin options or defaults) wins and the manual Toaster's props silently stop applying. No double-render (the registry guarantees that, as documented), but a position/duration config can visibly move. The spec covers the new model well; I'd add an explicit upgrade note: "if you pass props to a manual `<Toaster>` while also installing the plugin, move them to `app.use(ToastPlugin, {...})`."

robsongajunior (approved):

isaquebock (approved):

PR #817, opened 2026-07-28
fix(theme): align text tokens with typography primitives

gabriel-lisboa-azion:
## Summary

`texts.data.js` had drifted away from the primitive scales it is supposed to compose from: font sizes, line heights, weights, letter-spacing and the `text-link` transition were hardcoded string literals duplicating (or nearly duplicating) values that already exist as primitives. This aligns the text bundles with those primitives and brings label weights onto the scale.

Three source commits, each with its regenerated `dist/`:

1. **`fix(theme): reference typography primitives in text tokens`** — swaps the literals for the primitives they duplicate (`fontSize.*`, `leading.none`, `fontWeight.normal`, `tracking.normal` / `tracking.widest`, `duration['fast-02']` + `curve['productive-entrance']`), and adds `leading.none` (`= 1`) to the leading scale — the value the code label tokens were already hardcoding.
2. **`fix(theme): set label text tokens to normal font weight`** — `text-label-lg` / `-md` / `-sm` go from `fontWeight.medium` (500) to `fontWeight.normal` (400).
3. **`fix(theme): emit the letter-spacing scale in rem`** — the six `tracking` steps move from `em` to `rem`, keeping their numbers, so the emitted CSS carries a single measure unit.

Plus a merge of `main` and two `test(storybook)` baseline commits — see Verification.

## Notes

**Three visual changes ship here — all intentional:**

- **Overline letter-spacing `0.08em` → `0.1em`.** `0.08em` appears nowhere else in the repo and has no entry on the `tracking` scale (nearest are `wider: 0.05em` and `widest: 0.1em`). Rather than add a primitive used by exactly one token family, the three `text-overline-*` tokens snap onto `tracking.widest`, at the cost of `+0.02em`.
- **`text-label-{lg,md,sm}` weight `500` → `400`.** Every consumer of `.text-label-lg` / `.text-label-md` / `.text-label-sm` renders one weight step lighter, matching the rest of the typography scale (uniformly 400).
- **Overline tracking `0.1em` → `0.1rem` (unit change).** Review flagged the emitted CSS mixing measure units: `--text-overline-xs-letter-spacing: 0.1em` sat next to `--text-label-md-font-size: 0.875rem`. The `tracking` scale was the only `em` left in `dist/v4/globals.css`, so all six steps move to `rem` with their numbers unchanged. The scale reaches consumers almost entirely through the eight derived `--text-*-letter-spacing` vars (raw `tracking-*` utilities are banned by the `tracking-raw` check), five of which are `0` and inert — so the three `text-overline-*` tokens carry the whole change: `0.1em` on a `0.75rem` overline was 1.2px, a fixed `0.1rem` is 1.6px. Note that `rem` tracking no longer scales with the type size, so `text-overline-md` keeps the same absolute tracking at its `sm` breakpoint (`0.875rem`) as at its base size.

Taken together, overline tracking moves `0.96px` → `1.2px` → `1.6px` across this PR.

Everything else is byte-identical in the emitted CSS: `0` → `0rem` is equivalent, and the `text-link` transition resolves to the same `all 110ms cubic-bezier(0.39, 0.57, 0.56, 1)` (it does not even appear in the `dist` diff).

New `--leading-none: 1` CSS variable is additive; no token was removed or renamed, so this is not breaking.

## Verification

- `pnpm build:tokens` regenerated `dist/v4/globals.{css,scss}` per commit, and is idempotent (a second run produces no diff), including on the merged tree.
- `grep -cE '[0-9]em;' dist/v4/globals.css` returns **0** — no `em` value remains anywhere in the emitted CSS.
- Prettier and ESLint clean on every changed source file.
- The first two commits sum to exactly the original working tree (verified by diffing the branch total against the pre-split patch).
- **Governance is green on the head commit** (22/22 checks). The visual suite exercised the tracking change both ways: `Storybook visual tests (shard 2/4)` **failed** on the merge commit against `main`'s stale baselines, then all four shards passed once the regenerated set landed.
- **Baselines** regenerated on ubuntu via the `app-storybook-generate-baseline` workflow (the committed baselines are linux-only — `process.platform` keys the snapshot dir). Two rounds: **22** captures for the `rem` change (the three `Overline` stories × 6 modes, plus the two `Dropdown` top/bottom-slot stories on mobile), then **92** after merging `main` (32 `inputtext`, 24 `fieldtext`, 36 `fieldselect`), since #782 / #779 baselines predate this branch's tokens.
- Merging `main` conflicted on **60 binary baselines** and no source files (`catalog.json` auto-merged). Neither side is correct for the merged tree, so all 60 were resolved to `main`'s version and the full 1945-file matrix was then regenerated from the merged tree — the committed set is derived, not hand-picked. `main` did not touch `packages/theme`.

## Follow-up (out of scope)

`packages/webkit/src/components/overline/overline.vue` stacks `tracking-tightest` on top of `text-overline-md`. There is **no `tightest` step** on the `tracking` scale, so no `--tracking-tightest` variable exists and Tailwind v4 emits nothing for that class — it overrides nothing, and the component has always rendered the token's own letter-spacing. This PR's baseline run confirms it: Overline's own captures shifted with the token change.

*(Corrected — an earlier version of this note said the class "overrides whatever these overline tokens emit". It does not.)*

The real finding is that `tracking-tightest` is dead markup that the `tracking-raw` ESLint check in `token-checks.js` is meant to ban and currently misses. Worth a separate look.


robsongajunior (changes requested):
In the file `packages/theme/dist/v4/globals.css` note the multiple measure unit.

Example:
```
--text-overline-xs-letter-spacing: 0.1em;
--text-label-md-font-size: 0.875rem;
```

Note using `rem` and `em`. Should be pick only one.
Unfortunately passed in some pull request but didnt.

robsongajunior (approved):

isaquebock (approved):

PR #822, opened 2026-07-28
feat(webkit-sample): add a layout token system, equal-height Home columns and column-selector filters

gabriel-lisboa-azion:
## Summary

`apps/webkit-sample` had grown to ~34 pages that each invented their own content container. The four archetypes disagreed on all three axes that matter — width, boundary inset, and section rhythm — so the boundary read tight, sections sat too close together, and Home's two columns ended at different heights.

This encodes the container system **once, as CSS layout tokens**, and moves every archetype onto it. A token file is what makes the standard enforceable instead of aspirational: retuning the boundary is now a one-line edit rather than a 25-file sweep. It is shaped deliberately like theme tokens (semantic names, zero raw values) so the set can be promoted into `@aziontech/theme` as the first Azion **layout** token set — the promotion path is recorded in the file header.

Separately, the Applications and Workloads tables shipped `Table.Filter`, a generic field/operator/value popover builder. That is the wrong model for a console list; what a user wants is a selector per column.

Four commits, reviewable in isolation:

1. **`feat`** — the token layer (`src/styles/layout.css` + wiring, no page edits)
2. **`refactor`** — all four archetypes onto the tokens (AppLayout alone moves 19 padded pages)
3. **`style`** — Home's equal-height columns + a more generous resources empty state
4. **`feat`** — column-selector filters, and the shared `lib/dates.js` / `lib/filters.js` behind them

## Notes

**No new dependencies. No breaking changes.** `webkit-sample` is private and outside `release-please-config.json`'s paths, so none of these commits cuts a release.

**Verified in a real browser** (Playwright over the dev server), not just by build:

| check | result |
|---|---|
| All 43 routes | zero console / page errors |
| Home columns end at same `y` | delta `0` at 1600 and 1280, for both the 3-action (Workloads) and 2-action (Edge DNS) empty state |
| Archetype-C alignment | form column and save bar share both edges (320…1600) — the 320px save-bar misalignment is gone |
| Archetype-B alignment | tab bar aligns with the header breadcrumb (264/264); content is inset by the boundary (468) |
| Column selectors | presets narrow 12→7 rows; multi-author narrows and restores; filtering from page 2 rewinds to page 1; Clear + Apply restores |
| Token layer is wired | overriding `--layout-boundary-inline` at runtime moved every page's inset at once (card left 273→313px) |

### Three deliberate divergences from the original plan

- **The tab bar is chrome, not content.** It keeps its own `--spacing-md` edge padding and stays full-bleed, aligned with the header's breadcrumb, rather than sharing the content column's boundary. Tabs and content are therefore intentionally offset by 24px.
- **`period` and `:presets` do not compose** on `Calendar` — `isTwoPart` is `hasPresets && !period`, so `period` *replaces* the presets rail and the grid and swaps the placeholder for the literal "Period". Last Modified uses `:presets` (rail + custom range + the real placeholder). `clearable` is inert in the two-part trigger for the same reason, so reset is `<Calendar.Clear>` composed into the panel's `#footer`.
- **The `sm`→`lg` group-gap sweep covers 46 sites, not all 70.** The 24 inside drawer panels keep `sm`: their enclosing fieldset separates sections with the same `lg` value, so raising both would flatten the hierarchy.

### Two webkit fixes this surfaced (out of scope here)

- `table-search.vue` renders `InputText` with `v-bind="$attrs"` *before* a literal `size="small"`, so a consumer `size` always loses. Both `Table.Search` call sites dropped their ignored `size="large"`.
- `select.vue` declares `w-full` in its own static `class` alongside `:class="attrs.class"`, so a consumer `w-[…]` loses the specificity tie — the first toolbar render stacked into four rows. Worked around with a wrapper div; that root should merge through `cn`.

### Known, pre-existing, not fixed here

At a ~700px viewport the Applications / Workloads pagination row sits below the fold (bottom 769px). It was **already** clipped before this change (713px) — the boundary bump widens an existing overflow. `07c4f649` and `ce8e785f` tuned that `min-h-0` behaviour on purpose, so it is left alone as a follow-up; the row is still reachable by scrolling.

### Worth a second look in review

1. **The `lg` group gap.** The section-label→CardBox relationship goes 12px → 16/24px in 46 places. At 24px a small "General" label starts to float away from the card it titles — the value most worth an opinion, and the one the token file makes trivial to retune.
2. **Home's 48px gutter at ≥1280px** — 320 aside + 48 gutter + 912 section. Fine on paper; judge whether it reads as separation or as a hole.
3. **Content vs. chrome alignment.** `GlobalHeader` is a fixed `px-md` (16px), so at ≥640px content insets 24px while the breadcrumb stays at 16 — an 8px offset on every padded page, not fixable from the sample. Flow pages already have exactly this relationship and read fine, so I recommend accepting it; the alternative is `--layout-boundary-inline: var(--spacing-md)`, a one-line change in the token file.

### DIFF
--- a/apps/webkit-sample/src/style.css
+++ b/apps/webkit-sample/src/style.css
@@ -1,6 +1,8 @@
 /* Theme tokens + the Tailwind v4 entry are imported from src/main.js
    (`import '@aziontech/theme'`); see the note there. This file carries only
-   the app's own font faces, icon fonts and shell rules. */
+   the app's own layout tokens, font faces, icon fonts and shell rules. */
+/* Layout tokens — the container system (boundary / rhythm / measure). */
+@import './styles/layout.css';
 /* PrimeIcons — webkit components reference `pi pi-*` glyphs. */
 @import 'primeicons/primeicons.css';
 /* Azion icon library — monochrome brand/UI glyphs (`ai ai-*`, colored via
--- a/apps/webkit-sample/src/components/AccountSettings.vue
+++ b/apps/webkit-sample/src/components/AccountSettings.vue
@@ -511,14 +511,12 @@ const onTeamAction = (event, value, row) => {
 
       <!-- Active tab body: only this region scrolls, between the bar and the
            (Account Settings) Save bar pinned below. -->
-      <div
-        class="flex min-h-0 w-full flex-1 flex-col gap-[var(--spacing-lg)] overflow-auto p-[var(--spacing-md)]"
-      >
+      <div class="min-h-0 flex-1 overflow-auto">
         <!-- Tab: Account Settings — the account identity form. One flag locks
              every control while the request is in flight. -->
         <form
           v-if="activeTab === 'account-settings'"
-          class="flex min-w-0 flex-col gap-[var(--spacing-lg)]"
+          class="layout-column layout-boundary flex min-w-0 flex-col gap-[var(--layout-section-gap)]"
           aria-label="Account settings"
           novalidate
           @submit.prevent="submit"
@@ -535,7 +533,7 @@ const onTeamAction = (event, value, row) => {
             <legend class="sr-only">Account settings</legend>
 
             <!-- Section: General -->
-            <section class="flex flex-col gap-[var(--spacing-sm)]">
+            <section class="flex flex-col gap-[var(--layout-group-gap)]">
               <p class="px-[var(--spacing-xs)] text-heading-xxs text-[var(--text-default)]">
                 General
               </p>
@@ -582,7 +580,7 @@ const onTeamAction = (event, value, row) => {
             </section>
 
             <!-- Section: Company Information -->
-            <section class="flex flex-col gap-[var(--spacing-sm)]">
+            <section class="flex flex-col gap-[var(--layout-group-gap)]">
               <p class="px-[var(--spacing-xs)] text-heading-xxs text-[var(--text-default)]">
                 Company Information
               </p>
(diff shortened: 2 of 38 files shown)


apps/webkit-sample/src/style.css
```diff
@@ -1,6 +1,8 @@
 /* Theme tokens + the Tailwind v4 entry are imported from src/main.js
    (`import '@aziontech/theme'`); see the note there. This file carries only
-   the app's own font faces, icon fonts and shell rules. */
+   the app's own layout tokens, font faces, icon fonts and shell rules. */
+/* Layout tokens — the container system (boundary / rhythm / measure). */
+@import './styles/layout.css';
 /* PrimeIcons — webkit components reference `pi pi-*` glyphs. */
 @import 'primeicons/primeicons.css';
```
robsongajunior, line 7: de onde saiu isso `@import 'primeicons/primeicons.css';`?
se está vindo dai está sendo instalado no package.json? se estiver está errado pq vem o `@aziontech/icons`

robsongajunior (changes requested):
atualizar de `apps/webkit-samples` para `apps/samples/vite` ou `apps/samples/astro` algo assim para que possamos ter mais de uma sample de forma faci

apps/webkit-sample/src/style.css
gabriel-lisboa-azion: Cara, nem era para rolar isso... nem sei pq ele abriu isso aqui de fato. Eu não peço para abrir PR na branch da sample.... wtf

PR #830, opened 2026-07-29
refactor fix: forbid zero with a length unit across tokens, lint and docs

isaquebock:
## Summary

A zero length is identical in every unit, so the unit is pure noise — and it made the same zero read three different ways across the codebase (`0px`, `0rem`, `0em`). This forbids zero with a **length** unit and gates it on every authoring surface, so no path escapes it.

- **Rule** — `.claude/rules/styling.md` gains the "A zero length carries no unit" section (length units only; `0%`, `0s`/`0ms`, `0deg`, `0fr` are untouched). Mirrored into the shipped `webkit-tokens.md` and documented in `packages/webkit/docs/DOC_LINTS.md`.
- **Token checks** — `zero-with-unit` added to the shared engine in `token-checks.js`, so it blocks write-time (hook) and in the `check-authoring` CI ratchet over component sources.
- **Stylelint** — `length-zero-no-unit` enabled in `.stylelintrc.json` and in the shipped `stylelint-config.js`, set to plain `true` so the preset's `ignore: ['custom-properties']` does *not* apply (a design system is authored almost entirely as custom properties). A `lint-canaries/stylelint/length-zero-no-unit.css` fixture keeps that from being silently relaxed.
- **Theme tokens** — `tracking.normal` emits `0` instead of `0em`, with a build-time assertion in `build:tokens`. Token values are authored in JS and compiled, so `build:tokens` is the only gate that sees them.
- **Fixes** — the two existing violations (`accordion-content.vue`, `toaster.vue`).

## Notes

- No new dependencies. No breaking change — bump is **patch** for `theme` and `webkit`.
- The `.claude/rules` + skills edits ship **with** the code on purpose: `packages/webkit/test/standards/invariant.test.mjs` pairs each rule to its gates, so splitting the doc from the enforcement would fail CI on whichever side landed first.
- `packages/theme/dist/v4/globals.{css,scss}` changes are the regenerated build output of the `tracking.normal` token.

## The one carve-out: a zero inside a math function

Inside `calc()` / `min()` / `max()` / `clamp()` a bare `0` is a **number**, not a length, so CSS requires the unit and the whole expression is invalid without it. Stylelint's `length-zero-no-unit` skips math functions for exactly that reason — which means it accepts `calc(100% - 0px)`.

So where the unit is unavoidable, it is **`rem`** (the one length unit this system scales in), and the two engines we own enforce it: a new `zero-unit-in-calc` token check and the same assertion in `build:tokens` flag any non-`rem` unit on a zero inside a math function, while `zero-with-unit` stops reporting the zeros that legitimately keep a unit there. A test pins stylelint's skip so the split of responsibility stays visible.


packages/theme/src/scripts/build-tokens.mjs
```diff
@@ -357,6 +357,36 @@ const emitCssV4 = () => {
...
+  }
+  if (inMath.length > 0) {
+    parts.push(
+      `Inside calc()/min()/max()/clamp() the zero needs a unit, and that unit is rem: write '0rem'.\n${detail(inMath)}`,
+    );
+  }
+  parts.push('Fix the token source under src/tokens/, not the generated CSS.');
+  throw new Error(parts.join('\n'));
+};
+
```
robsongajunior, line 396: this function should be removed after @isaquebock activate the running lint in the @aziontech/theme

robsongajunior (approved):

robsongajunior (changes requested):
pr está poluído. nao sei onde termina o review

packages/theme/src/scripts/build-tokens.mjs
isaquebock: agree

robsongajunior (approved):

paulo-sobrinho-azion (approved):

PR #837, opened 2026-07-30
fix(webkit): [ENG-47034] truncate the breadcrumb on one row and pin every glyph size

herbert-julio-azion:
Closes ENG-47034.

## What was wrong

Three separate defects, each measured before and after.

**The trail wrapped instead of truncating.** The shrink chain reached the anchor but stopped
there: the item root was `inline-flex`, and an inline-level box resolves its width
shrink-to-fit against its own max-content, so inside a hand-composed `<li>` (which arrives
with no class, hence a block) the anchor stayed at its text width and spilled out of the
item the row had already shrunk. Measured a 134px anchor inside a 92px item, label never
ellipsizing. The root is now a block-level `flex`, so it takes the width its parent gives
it, and the label carries `min-w-0` as the last link in the chain. Measured after: the
composed trail fits 375/375 with every label ellipsized.

**The glyphs changed size outside the design system.** Measured 14x14 at both 375px and
1280px, so they were already fixed inside webkit; in a consuming app
`.azion .pi { font-size: 1rem }` is unlayered CSS and beats anything in `@layer utilities`.
Every glyph is now pinned with `!`.

**The overflow trigger sat on top of the chevron.** It used `px-1` where a breadcrumb item
uses `px-[var(--spacing-xs)]`, leaving 4px between the ellipsis and the next chevron. It
now matches an item.

**The long label truncated with room to spare.** That was the story, not the component: it
hardcoded `class="max-w-[20rem]"`, a 320px cap that ignores the space available. Measured
without it: 1440px and 800px show the label in full (560/560), 375px ellipsizes (281/560).

Note that truncation is driven by the **space available**, not by a breakpoint: it is
`flex-shrink` against whatever width the parent gives. Only the collapse into the overflow
menu is viewport-driven.

## Stories

The composed and mobile-viewport stories are gone: the set documents the versions and the
mobile viewport is exercised by hand. The collapse keeps its coverage, rendered directly
instead of through the removed story, and the 18 orphaned baselines are pruned.

## Verified

28 tests pass (1 pre-existing skip), lint clean, the 89 story files comply with
`storybook-source.md`, spec and checksum updated.


packages/webkit/src/components/navigation/breadcrumb-item/breadcrumb-item.vue
```diff
@@ -73,7 +73,16 @@
...
-          'group/breadcrumb-item relative inline-flex min-h-6 shrink-0 items-center justify-center',
+          // Block-level `flex`, not `inline-flex`: an inline-level box resolves its
+          // width shrink-to-fit against its own max-content, so inside a
+          // hand-composed `<li>` (which arrives with no class, hence a block) the
+          // anchor stayed at its text width and spilled out of the item the row had
+          // already shrunk — measured 134px of anchor inside a 92px `<li>`, with the
+          // label never ellipsizing. A block-level flex box takes the width its
+          // parent gives it, which is what lets the label below actually truncate.
+          // In the data-driven mode the `<li>` is a flex container, so its child is
+          // blockified either way and nothing changes there.
```
robsongajunior: remove the comments

packages/webkit/src/components/navigation/breadcrumb-item/breadcrumb-item.vue
```diff
@@ -98,12 +107,17 @@
...
-      class="relative z-[1] size-[0.875rem] shrink-0 text-[length:inherit] leading-none"
+      class="relative z-[1] size-[0.875rem]! shrink-0 text-[0.875rem]! leading-none"
       aria-hidden="true"
       :data-testid="`${testId}__icon`"
     />
+    <!-- `min-w-0` is the last link in the shrink chain. Without it the span keeps
+         `min-width: auto`, refuses to go below its text width, and overflows the
+         anchor instead of ellipsizing: measured 118px of text inside a 92px item,
+         with no ellipsis. The list and the anchor already allow shrinking; this is
+         where it stopped. -->
```
robsongajunior: remove it

packages/webkit/src/components/navigation/breadcrumb/composables/use-breadcrumb-collapse.ts
```diff
@@ -0,0 +1,21 @@
...
+export const breadcrumbCollapseMediaQuery = '(max-width: 767px)'
+
+/**
+ * True while the trail should render its collapsed form.
+ *
+ * The breadcrumb owns this instead of borrowing the overlay layer's mobile
+ * composable: it is a navigation concern, and keeping it here is what lets the
+ * rule evolve (for example toward measuring the space actually available rather
+ * than the viewport) without touching how overlays decide to become sheets.
+ */
```
robsongajunior: remove it

packages/webkit/src/components/navigation/breadcrumb/composables/use-breadcrumb-collapse.ts
```diff
@@ -0,0 +1,21 @@
+import { useMediaQuery } from '@vueuse/core'
+import { readonly, type Ref } from 'vue'
+
+/**
+ * Width below which the trail collapses its middle segments into an overflow
+ * menu. Matches Tailwind `max-md` (below `breakpoints.md` / 768px), so the
+ * JS-driven collapse and the CSS-driven layout switch at the same point.
+ */
```
robsongajunior: remove it

robsongajunior (dismissed):

packages/webkit/src/components/navigation/breadcrumb-item/breadcrumb-item.vue
HerbertJulio: Resolved in c0c78717: the 9-line block above the root class list is gone. The behavioural part of it (block-level `flex` instead of `inline-flex`, plus `min-w-0`) stays exactly as it was; only the prose was removed.

packages/webkit/src/components/navigation/breadcrumb-item/breadcrumb-item.vue
HerbertJulio: Resolved in c0c78717: the HTML comment above the label span is gone. `min-w-0 truncate` on the span is unchanged.

packages/webkit/src/components/navigation/breadcrumb/composables/use-breadcrumb-collapse.ts
HerbertJulio: Resolved in c0c78717: the JSDoc block above `useBreadcrumbCollapse` is gone. The composable itself (and its explicit `Readonly<Ref<boolean>>` return type) is unchanged.

packages/webkit/src/components/navigation/breadcrumb/composables/use-breadcrumb-collapse.ts
HerbertJulio: Resolved in c0c78717: the JSDoc block above `breadcrumbCollapseMediaQuery` is gone. The `(max-width: 767px)` value is unchanged.

robsongajunior (approved):

guilherme-santana-azion (approved):

PR #838, opened 2026-07-31
feat(webkit): add show-fields to the calendar panel and cap the panel width

gabriel-lisboa-azion:
## What

Adds a `show-fields` prop to `Calendar` (default `true`) and caps the popover panel width.

- `show-fields: false` suppresses the Start/End date fields, leaving the grid, the presets rail and Apply. The interaction some consoles want is "pick from shortcuts or click the grid" — typing a boundary by hand is not offered there, and an empty pair of fields is dead weight in the panel.
- The panel is content-sized, so its **widest child** sets its width — and one of those children is the consumer's `#footer` slot. A row of shortcut chips stretched the panel without bound and spread the month grid across it. `max-w-[min(90vw,44rem)]` is the hard stop: `rem` (not a px container token) because the content that grows is text, so the cap grows with the root font the same way the grid and the chips do; `90vw` keeps it inside a narrow viewport. It sits above every built-in layout, including `horizontal`, so nothing that ships is clipped.

## Why

Both came out of building a console screen that drives the calendar from shortcut chips in the footer: the fields were redundant there, and the chips blew the panel open.

## Surface

| | |
|---|---|
| Props | `+ showFields?: boolean = true` |
| Events / slots | unchanged |
| Spec | `.specs/calendar.md` — Purpose + Props row, checksum refreshed |
| Catalog | `catalog.json` rebuilt (`catalog:build`) so `catalog:check` stays green |

## Tests

Two cases added to `calendar.test.ts`: `show-fields=false` hides `input-calendar__fields` while keeping the grid and Apply; the default renders the fields. Existing suite unchanged.

## Note

`calendar.vue` and `calendar.test.ts` are also touched by **#827** (`fix/calendar-field-parity`), which adds neither of these. The two are independent but will conflict textually — whichever lands second rebases.


herbert-julio-azion (approved):

robsongajunior (approved):

guilherme-santana-azion (approved):

packages/webkit/src/components/inputs/calendar/calendar.vue
```diff
@@ -635,6 +638,14 @@
     </span>
 
     <Teleport to="body">
+      <!-- The panel is content-sized, so its WIDEST child sets its width — and one of
```
isaque-bock-azion, line 641: Remover esses comentários.

isaque-bock-azion (approved):

robsongajunior (approved):

PR #843, opened 2026-07-31
fix(webkit): keep switch chrome intact inside an input-group

gabriel-lisboa-azion:
## What

`InputGroup`'s button-flattening selectors now skip a `Switch`.

```diff
-[&_button]:!border-transparent [&_button]:!rounded-none
+[&_button:not([role=switch])]:!border-transparent [&_button:not([role=switch])]:!rounded-none
```

## Why

`InputGroup` deliberately flattens the chrome of the controls it wraps — transparent borders and square corners — so a group of segments reads as one field with internal dividers. That is right for a trailing `Button`, an `IconButton` or a combobox trigger.

A `Switch` is also a `<button>`, but it is **not a segment of the field**: it renders a pill (`rounded-[var(--shape-elements)]`, `h-5 w-9`, `border-[var(--border-default)]`, and `data-[checked]:border-transparent` for its own checked state). The blanket `[&_button]` selectors squared its corners and erased its border with `!important`, so a switch placed inside a group rendered as a small grey rectangle with a floating thumb — the control still worked but no longer looked like a switch.

Scoping the two selectors with `:not([role=switch])` (the role `switch.vue` already sets on its root) leaves every other button flattened exactly as before and hands the switch its own chrome back.

## Scope

Two selectors on one class string. No props, events, slots or tokens change, and no other control matches `[role=switch]`.


herbert-julio-azion (dismissed):

guilherme-santana-azion (changes requested):
image

image

Quando o switch esta desabilitado o toogle respeita o padding left, porem quando esta habilitado não respeita o padding right. Isso causa um ""padding"" diferente entre ativo e desativado

guilherme-santana-azion (dismissed):

isaque-bock-azion (approved):

robsongajunior (approved):

PR #850, opened 2026-08-04
fix(theme): [ENG-46996] harden the zero-with-unit token gate

isaque-bock-azion:
## Summary

Hardens the zero-with-unit discipline that guards the compiled token stylesheet (`packages/theme`) and its mirror in the shared webkit token-checks engine. Addresses ENG-46996.

- **Complete unit list** — adds the CSS length units a partial list dropped: logical viewport (`vi`/`vb`), their small/large/dynamic + min/max variants (`svi`/`svb`/`svmin`/`svmax`/`lvi`/`lvb`/`lvmin`/`lvmax`/`dvi`/`dvb`/`dvmin`/`dvmax`), and the root-relative font units (`rex`/`rch`/`ric`/`rcap`). `0vi` no longer slips through.
- **Case-insensitive** — already in place (`i`/`gi` flag); now pinned by tests (`0PX`, `0Rem`, `0VI`, `0SvB`).
- **No longer one-declaration-per-line** — the assert scans the generated stylesheet as a whole string instead of `split('\n')`, so a `calc()`/`min()`/`max()`/`clamp()` wrapped across lines is judged in context (the math-function lookbehind spans the newline) rather than misread as a bare zero. No PostCSS — the theme's node-built-ins-only constraint is preserved.
- **Honest error path** — the reported filename is a parameter now, not a hardcoded `globals.css`.
- **Actionable error** — the message names the offending custom property so the source token is greppable under `src/tokens/**`.
- **Testable** — the pure logic moved to `packages/theme/src/scripts/zero-unit.mjs` (no side effects) so it is unit-testable without running the build. `build-tokens.mjs` imports `assertNoZeroWithUnit`. A theme `test` script runs the suite.
- **Kept in sync** — the shared webkit `token-checks.js` engine gets the identical completed unit list (both zero regexes now carry the same string); both copies are pinned by tests.

## Notes

- **Implementation note preserved:** the module regexes stay non-global so `.test()` is stateless; `findZeroMisuse` clones with the `g` flag locally to enumerate matches.
- **Item "avaliar checar a fonte":** the error names the offending token (greppable in `src/tokens/**`); full provenance to the exact source file was evaluated and deferred as disproportionate to the single-pass, dependency-free build.
- **CI wiring deferred:** a `theme-tests` job for `governance.yml` was prepared but left out of this PR (the push token lacks the `workflow` scope). Until it lands, the theme suite runs via `pnpm --filter @aziontech/theme test`; the webkit copy is already gated by the existing `toolkit` job. To be added in a follow-up with a `workflow`-scoped push.
- **Verified:** theme suite 12/12, webkit `token-checks` 9/9, `build:tokens` end-to-end clean; the widened list catches no existing component source. Release effect: `fix` → patch.
- Touches the same regexes as ENG-47001 (canonical-syntax work, not yet merged) — expect a rebase once that lands.


isaque-bock-azion:
Revisei e enxuguei os comentários dos arquivos deste PR em `packages/theme` (`src/scripts/zero-unit.mjs` e `test/zero-unit.test.mjs`) — mais curtos e diretos, sem mudar lógica nem cobertura (theme 12/12, `build:tokens` limpo).

Escopei a revisão aos arquivos que **este PR adiciona/altera** no theme, para não misturar um sweep de comentários do pacote inteiro (ex.: o cabeçalho e as seções pré-existentes do `build-tokens.mjs`, que não fazem parte desta mudança) e evitar diff não-relacionado. Se quiser, faço uma passada separada nos comentários do resto do `packages/theme` num PR próprio.

guilherme-santana-azion (approved):

herbert-julio-azion (approved):

PR #852, opened 2026-08-04
docs: require signed commits and document the SSH signing setup

herbert-julio-azion:
## Summary

PR #838 is approved, has all 23 checks green and no conflicts, and still cannot be merged. The cause is invisible from inside the PR: the `main` ruleset (`Protected branchs`) gained `required_signatures` on 2026-07-30, and the branch commit is unsigned, so the merge state is `BLOCKED` with no failing gate to point at.

Nothing in the repo documented that requirement, and no machine on the team signs commits today: every locally authored commit across the open PRs is `unsigned`, and the only verified commits are the ones GitHub itself generates (web UI edits, the "Update branch" button, the squash commit on merge). This adds the requirement to the two places contributors already read.

## Notes

- `CONTRIBUTING.md` gains a `### Signed commits` section under Commit convention: the one time SSH signing setup, registering the key a second time as a **Signing Key** (an authentication key does not verify signatures), the macOS trap where git < 2.34 makes `gpg.format ssh` fail on every git command including `git status`, how to inspect signatures locally (`gpg.ssh.allowedSignersFile`, or the configuration free `gpgsig` header check), and how to recover a branch that already carries unsigned commits.
- `.claude/rules/git-workflow.md` gains a `## Signed commits` section with the invariant and the three non fixes: reopening the PR, re-pushing the same commits under a different author, and fixing one PR expecting the next one to work. It also records that re-signing rewrites history and therefore dismisses existing approvals (`dismiss_stale_reviews_on_push`).
- Enforcement is already server side (`required_signatures`), so this documents an existing blocking gate rather than adding an advisory one.
- Docs only: `docs` produces no version bump.
- This PR's own commit is signed, so it also serves as a check that the documented setup produces what the ruleset wants.


guilherme-santana-azion:
**Sugestão**

Não seria bom ter uma section falando sobre assinar o CLA?

robsongajunior (approved):

isaque-bock-azion (approved):

PR #854, opened 2026-08-04
feat(webkit): give Sidebar a resizable, collapsible rail

gabriel-lisboa-azion:
> **Stacked on #853** (`feat/menu-component`). It targets `main` so Governance actually runs — a PR targeting a non-`main` branch gets zero CI. Until #853 merges this diff carries its commit too; afterwards it collapses to the rail alone. **Review #853 first.**

The drag / snap / keyboard-nudge gesture and the collapse affordance were being re-implemented in every consuming shell — the console rail and the docs rail each had their own copy. They belong to the component.

## API

| Prop | Does |
|---|---|
| `resizable` | Adds the trailing-edge drag handle; dragging past the minimum collapses the rail. |
| `collapsible` | Adds the collapse trigger at the bottom of the rail, plus the edge affordance that brings a collapsed rail back. |
| `minWidthToken` / `maxWidthToken` | Theme container tokens the sized width is clamped to, read off the document at runtime (default `--container-3xs` … `--container-sm`). |
| `collapseAriaLabel` / `expandAriaLabel` / `resizeAriaLabel` | Accessible names for the three controls. |

Both states report through `v-model:collapsed` and `v-model:width` (the width already clamped), so the host owns persistence.

Width is clamped between **theme container tokens** rather than pixel props, so a host sizes the rail in the same scale as the rest of the system.

`measure()` is exposed for a host that reveals the sidebar *after* mount — e.g. a viewport change out of a mobile layout. A rail measured while `display: none` reports 0, which would otherwise strand it invisible.

## Where the pieces live

The gesture is `composables/use-sidebar-rail.ts`; its timings are `presets/transitions.ts`, so the motion comes from the `duration` / `curve` tokens instead of literals, and the reduced-motion path short-circuits to `transition: none` (the rail simply *is* where it lands). `sidebar.vue` also sets both `--menu-slide-surface` and `--menu-ring-offset` on its `<nav>`, so a `Menu` dropped inside rings against the sidebar's own fill with nothing to wire.

## Stories

`Sidebar.stories.js` moves onto `Menu`, which is what a new sidebar composes. That replaces `WithHeaderAndProfileFooter` and `WithHeaderSearch` with `Default` + `Resizable`, so **12 of its committed baselines are orphaned** and are removed in this branch's baseline commit — a regen artifact adds and updates files but never deletes.

`sidebar-group.test.ts` drops its story fixture for the same reason (the stories no longer render a `SidebarGroup`) and exercises the component directly. `SidebarGroup` itself is untouched and still supported; it is not deprecated here, because `webkit/no-deprecated-component` is `error` in the recommended config and would turn a consumer's lint green→red on a minor.

## Follow-up in this PR

Linux visual baselines still to be regenerated via CI (new `Resizable` story, plus the 12 orphan removals above).


gabriel-lisboa-azion:
Addressed the six CodeQL `property access on null or undefined` alerts in `menu.test.ts` (34af255d).

Root cause: CodeQL's JS extractor cannot resolve the `./menu.vue` import behind the compound root, so it reads `Menu` as `undefined` and flags every dot-notation assertion. The compound is correctly wired at runtime, which is exactly what the test proves.

Fix: assert the root is defined, then reach its members through optional chaining. The assertions keep their meaning, because `undefined` never equals a sub-component, so a member that goes missing still fails. `pnpm vitest src/components/navigation/menu/menu.test.ts` passes 29/29, and lint and `vue-tsc` are clean. Threads resolved.

isaque-bock-azion:
Bug de navegacao, quando o clico fora do Dialog de search, o search permanece em focus, precisando de 2 clicks para seleciona-lo novamente.
https://github.com/user-attachments/assets/6fc8a612-f691-42d2-ac4b-6ff6917c9d3b

isaque-bock-azion:
Nas stories do sidebar, os clicks nos items de menu redirecionam para um Not Found interno, recomendo nao redirecionar.

isaque-bock-azion:
Dúvida: deveria ter focus nessa area? 
image

Ela nao é acionavel

isaque-bock-azion:
image
Focus do primeiro item abaixo do search está cortado

isaque-bock-azion:
Dúvida: Está correto a forma como está mostrando esse botao de resize?
image

isaque-bock-azion:
Parece que está com double focus: 
image
image

isaque-bock-azion:
Focus interno no options do user está cortado

image

isaque-bock-azion:
Sugestao: Podemos dar um respiro ao divider nesse cenário.

image

image

gabriel-lisboa-azion:
### Review fixes — the sidebar's first row and the footer line

Three fixes in `f4b5e4c`, all verified in a real browser against the `Resizable` story (1280×720, light).

**1. The scroll viewport was eating the first tab stop.**
`ScrollArea` hard-coded `tabindex="0"`, so Tab out of the header search landed on the box around the navigation, not on the first row. A scroll region earns that stop only when nothing inside it is focusable, and this one holds a `Menu` of rows — so the sidebar passes `tabindex="-1"` and `ScrollArea` now reads a consumer-supplied `tabindex` instead of hard-coding it (the default stays `0`, so every other consumer is untouched, and the region is still programmatically focusable for scroll-into-view).

```
before → Tab lands on  DIV  [data-testid="layout-sidebar__scroll"]  tabindex="0"
after  → Tab lands on  A    "Home"
```

**2. The first row's focus ring was clipped.**
With a header the region ran `pt-0`, so the first row sat flush against the clip and lost the outer 4 px of its ring (`ring-2` over `ring-offset-2`). It now keeps `--spacing-xxs` there, and reserves the same as `scroll-padding` so a row the keyboard scrolls into view stops short of the edge rather than on it.

```
room above the first row:  0 px → 4 px      (ring extent is exactly 4 px)
```

**3. The collapse trigger was off the footer line, and the separator stopped short of it.**
`SidebarFooter` owned the band (`border-t` + `pt-md`) while the trigger sat outside it, so `items-center` centred the trigger against a box a padding taller than the content row, and the line ended before the trigger. The band moves to the sidebar's own footer region, which wraps both; `SidebarFooter` drops its own whenever it is inside a collapsible sidebar, so nothing doubles up and a non-collapsible footer is byte-identical.

| | separator span (of a 16 px-inset 392 px region) | avatar centre | ⋮ centre | « centre |
|---|---|---|---|---|
| before | `16 → 355` (37 px short) | 529.5 | 529.5 | **523.5** |
| after | `32 → 407` (full width) | 545.5 | 545.5 | **545.5** |

The spec's own rule — *"the footer content must not add its own top padding, which would drop it below the trigger it is meant to line up with"* — was being broken by `SidebarFooter` itself; § *Resizable and collapsible* now describes where the band lives.

Covered by three new assertions (viewport `tabindex`, Tab reaching the first row, and content + trigger sharing one band) — full suite green at 87 files / 1810 tests. Linux baselines regenerating.

gabriel-lisboa-azion:
### The collapsed rail now previews itself (`08fa095`)

A collapsed rail leaves nothing on the page, so the way back was a 2 px line you had to find and trust. Resting the pointer in the leading-edge zone — or landing focus in it — animates the rail back in to `--size-10` and morphs the page on the same frames.

Measured through the interaction (Chromium, 1280×720, `Resizable` collapsed):

| | rail width | zone width | `main` left |
|---|---|---|---|
| at rest | 0 | 24 (`--size-6`) | 16 |
| 90 ms after entering | 32.4 | 37.4 | 48.4 |
| settled | 40 (`--size-10`) | 40 | 56 |
| pointer deep inside the sliver | 40 | 40 | 56 |
| after leaving | 0 | 24 | 16 |

**Three decisions worth reviewing:**

**The zone grows with what it opened** (`--size-6` → `--size-10`). A fixed narrow zone ends where the sliver begins, and the pointer crossing that seam flickers the preview on and off. Sizing the zone to the sliver keeps it under the pointer — the "deep inside the sliver" row above is that case.

**Both widths come from the fixed `--size-*` scale, deliberately.** I first used `--spacing-xl` and measured the zone at **48 px**, not 24: `--spacing-lg`/`xl`/`xxl` are breakpoint-responsive in the theme. That is far more of the page's leading edge than a hit target should claim, and it silently changes per viewport. `--size-*` is flat.

**The sliver is a window, not a resize.** The panel keeps its committed width and the rail clips it, so nothing re-wraps and there is no second layout to design. Row content therefore meets the clip mid-glyph — icons sit at panel-local **28–44 px**, so a 40 px window cuts every one by 4 px whatever else we do. The panel's last `--size-9` → `--size-10` is masked to transparent so the cut reads as "there is more behind this edge" rather than as a rendering bug. Stops are absolute, not percentages: the mask box is the panel's own width, so `100%` would put the fade off-screen.

> If you'd rather the icon column land whole, the sliver wants `--size-11` (44 px) — 40 px cannot contain it given the region's 16 px padding plus the row's 12 px. Say the word and it's a one-token change.

A preview is **not** a restore: the rail stays `inert` + `aria-hidden` throughout, and the expand button is still what brings it back. That button also stopped swallowing clicks on the page's leading edge while invisible (`pointer-events-none` until shown).

Four new tests cover it (pointer opens/closes, focus does the same, the sliver stays out of the tab order, an expanded rail renders no zone). Full suite green at 87 files / 1814 tests; `vue-tsc`, lint, catalog, test-gate, authoring and doc-standards all pass.

**No baseline churn:** the resting render is byte-identical — I diffed the `Resizable` canvas before/after at 0 of 921 600 pixels. The preview is a hover state, and the zone only exists while collapsed.

isaque-bock-azion (dismissed):

herbert-julio-azion (dismissed):

robsongajunior (approved):

guilherme-santana-azion (approved):

PR #864, opened 2026-08-06
fix(webkit): make the NavigationMenu open out of its trigger and morph between panels

gabriel-lisboa-azion:
## Summary

- Every enter and exit in this component silently snapped: the transition lists named `transform`, but Tailwind v4's `scale-*` / `translate-*` utilities write the individual `scale:` / `translate:` properties, which naming `transform` does not cover. Only the positioner is a real `transform` (an inline `translate3d`) — and it was transitioning `top/left/right/bottom`, none of which it sets, so the panel **jumped** between triggers instead of travelling.
- The popup now scales out of `--popup-origin`, the active trigger, so opening reads as the menu coming out of the thing that was pressed; switching panels is one morph, with box size and placement sharing a duration and curve while the two panels cross-fade over a shorter one.
- The content slide was `translate-x-1/2` — half the panel width, ~500px on a wide mega-menu. It is now a 24px directional nudge keyed on the side the new item sits on.

## How to test

1. Open `Components/Navigation/NavigationMenu` in Storybook and click a trigger: the popup should scale up **out of that trigger**, not slide down from above it.
2. Move to a sibling trigger without closing: the box should travel and reshape as one object, and the panel underneath should cross-fade with a short slide from the side the new item sits on. Nothing should jump.
3. Close it: the panel should scale and fade back out rather than disappearing.
4. Re-open on a **narrower** window so a panel would re-wrap: the panel must keep its own width through the morph — the box crops it, it does not re-flow mid-animation.
5. Enable OS "reduce motion" and repeat: every transition above should be off, with no snapping into a broken position.
6. `pnpm --filter @aziontech/webkit exec vitest run src/components/navigation/navigation-menu` — 16 tests pass.

## Notes

- **Depends on `fix/theme-named-motion-utilities`** (pushed, no PR yet). `duration-moderate-02` / `ease-productive-entrance` currently compile to nothing because the `--transition-duration-*` / `--ease-*` namespaces are never emitted, so until that lands the timings here fall back to Tailwind's default 150ms `ease`. The motion is correct either way; the *curves and durations* need that branch. This PR does not carry it, because it changes motion timing for 46 files across the system and deserves its own review.
- `use-navigation-menu-viewport-size` becomes `.ts` for the panel measurement, which also clears its `composable-js` entry from `authoring-baseline.json`.
- No new dependency, no breaking change. Positioning and animation stay CSS-only.
- Linux visual baselines may need regenerating if any open-menu screenshot shifts — the panel sizing path changed.


robsongajunior (changes requested):
a lot of unuseful comments, remove all and keep only for props

robsongajunior (approved):

guilherme-santana-azion (approved):

